PR #2, opened 2018-10-29
Installation updates

thomasbazeille:
Updated setup.py and version.py arguments copied too quickly. Library now installs smoothly.


bthirion:
LGTM. Merging.

Issue #7, opened 2019-02-07
Implement a diagonal transformation, projecting X on Y through voxel signal scaling


thomasbazeille:

bthirion:
I was thinking that scaling could be done after any estimator, though for some, such as ridge it makes no sense.

bthirion:
A related question is: compare rescaled ortogonal transform with bootstraraped orthogonal transform: if bootstrap brings nothing more than scaling, then it is essentially useless !

thomasbazeille:
> I was thinking that scaling could be done after any estimator, though for some, such as ridge it makes no sense.

You mean, first a transform (for example orthogonal, or a permutation) and then a voxelwise projection of the prediction on the target ? I fear if both are learnt consecutively on the train set, it will just foster overfit. This also makes our models more complex. 

What I propose is first to make projection available as a separate estimator to see how much the shrinkage of the signal it induces contributes to improve the prediction metrics and afterwards if needed to make it available after every model.

In the meantime, it will always be possible to chain the estimators if you need it for example :
`orthogonal.fit(X, Y)
X_transformed = orthogonal.transform(X)
diagonal.fit(X_transformed, Y)
X_transformed_rescaled = diagonal.transform(orthogonal.transform(X))`

bthirion:
Well it makes sense for methods like permutation.
However, i like the idea of chaining estimators explicitly.

PR #4, opened 2019-02-06
[MRG] Pairwise alignment module, tests and examples

thomasbazeille:

PR #10, opened 2019-02-13
[MRG] Implemented diagonal estimator to solve Issue #7

thomasbazeille:
I added/modified tests for projection scheme and class appropriatly.
New tests pass except once that can fail from time to time but this problem is already patched in PR #4

### DIFF
--- a/fmralign/alignment_methods.py
+++ b/fmralign/alignment_methods.py
@@ -83,12 +83,38 @@ def optimal_permutation(X, Y):
     return permutation
 
 
-def projection(x, y):
+def _projection(x, y):
+    """Compute scalar d minimizing ||dx-y||
+    Parameters
+    ----------
+    x: (n_features) nd array
+        source vector
+    y: (n_features) nd array
+        target vector
+
+    Returns
+    ----------
+    d: int
+        scaling factor
+    """
     return np.dot(x, y) / np.linalg.norm(x)**2
 
 
-def voxelwise_signal_projection(X, Y, n_jobs):
-    return Parallel(n_jobs)(delayed(projection)(
+def _voxelwise_signal_projection(X, Y, n_jobs=1):
+    """Compute D, list of scalar d_i minimizing ||d_i x_i-y_i|| for every x_i,y_i in X,Y
+    Parameters
+    ----------
+    X: (n_samples, n_features) nd array
+        source data
+    Y: (n_samples, n_features) nd array
+        target data
+
+    Returns
+    ----------
+    D: list of ints
+        List of optimal scaling factors
+    """
+    return Parallel(n_jobs)(delayed(_projection)(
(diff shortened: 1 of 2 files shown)


fmralign/alignment_methods.py
```diff
@@ -82,6 +83,16 @@ def optimal_permutation(X, Y):
     return permutation
 
 
+def projection(x, y):
+    return np.dot(x, y) / np.linalg.norm(x)**2
```
bthirion: docstring ?

fmralign/alignment_methods.py
```diff
@@ -82,6 +83,16 @@ def optimal_permutation(X, Y):
...
+
+
+def voxelwise_signal_projection(X, Y, n_jobs):
+    return Parallel(n_jobs)(delayed(projection)(
```
bthirion: dosctring here too. Also, as it lay not be part of the public API, you may prepend an undrescore:
_voxelwise_signal_projection
Same for the previous function.

fmralign/tests/test_alignment_methods.py
```diff
@@ -1,7 +1,7 @@
 import numpy as np
 from sklearn.utils.testing import assert_array_almost_equal, assert_greater
 from scipy.linalg import orthogonal_procrustes
-from fmralign.alignment_methods import scaled_procrustes, optimal_permutation, Identity, ScaledOrthogonalAlignment, Hungarian, RidgeAlignment, OptimalTransportAlignment
+from fmralign.alignment_methods import scaled_procrustes, optimal_permutation, voxelwise_signal_projection, Identity, DiagonalAlignment, ScaledOrthogonalAlignment, Hungarian, RidgeAlignment, OptimalTransportAlignment
```
bthirion: pep8: line too long.

fmralign/alignment_methods.py
bthirion: Actually, I'm wondering: you want to directly apply that at the matrix level, avoiding the loop.

thomasbazeille: After doing the benchmark, this is 4x faster this way:
```
from fmralign.alignment_methods import _voxelwise_signal_projection
import numpy as np
X = np.random.randn(100, 500000)
Y = np.random.randn(100, 500000)
```


`%timeit -n 100 _voxelwise_signal_projection(X, Y, 1)`
> 100 loops, best of 3: 109 ms per loop

`%timeit - n 100 np.sum(X * Y, 1) / np.sum(X * X, 1)`
 > 100 loops, best of 3: 703 ms per loop

thomasbazeille:
Good to merge

bthirion:
Amazing. OK, I owe you a beer ;-) 
B

bthirion:
Thx, merging

PR #21, opened 2019-03-14
[MRG] Pass parallelization back-end as an argument

thomasbazeille:
Convenience change to be able to run the library in different types of clusters (here dask for margaret). Default behavior is not changed but more flexibility is added


thomasbazeille:
Added a unittest for another back-end please review/merge when possible

thomasbazeille:
Ready for reveiwing @bthirion

bthirion:
LGTM. merging

Issue #18, opened 2019-02-26
make POT and optional dependency


bthirion:
Because it is harder to install (e.g. fails with cython 0.29). I had to upgrade cython to get it....

thomasbazeille:
The problem seem to occur specifically in anaconda, the quickfix is to run
`conda install --upgrade cython`

It is related to this issue : https://github.com/rflamary/POT/issues/59 but when I tried replicating the problem in a fresh environment I was unable to reproduce it, `pip install fmralign` was running smoothly.

bthirion:
Yep, but users son't always have a fresh environment, and they may really be bothered by such errors (that are fairly unfriendly to non-experts). This is the reason why Nilearn has no dependency on Cython !
Making POT an optional dependency is not complicated I think, and you can probably get most of the hyperalignment benefits with cheaper methods than POT.

agramfort:
what part of POT do you need? maybe you can just take the plain numpy code
to avoid the strong dep on a package that is pretty bleeding edge...

>

thomasbazeille:
The major method we use for now is the log-stabilized sinkhorn, you right maybe we don't need to have the full dependency for that. I did not consider taking the relevant code to not duplicate test and maintenance but maybe it is less costly after all than being dependent on POT.

Will make the dependence optional for the time being.

thomasbazeille:
Solved by #23

PR #23, opened 2019-03-14
Make POT an optional dependency

thomasbazeille:
to solve #18


thomasbazeille:
Ready for review.

bthirion:
We really need to setup travis, to see whether PRs break anything, as currently we're working in the dark.
So please, don't update any more the library without travis validation !

thomasbazeille:
Actually I have CircleCI running the test on my fork at every push and
nothing is broken. you should set it up as well for the Parietal account
and this project and the config.yml will take care lf the rest. We can do
that next week.

thomasbazeille:
@bthirion cf the comment above

bthirion:
Merging.

Issue #22, opened 2019-03-14
Improve parallelization


thomasbazeille:
PairwiseAlignment.transform() method could use some parallelization

thomasbazeille:
And the parallel back-end argument should be removed, to be replaced by just a "preferred back-end" that can be over-riden if necessary

thomasbazeille:
Solved by #32

Issue #19, opened 2019-03-05
Input list of lists of Niimgs in TemplateAlignment class


thomasbazeille:
For now, it only works properly with lists of 4D images, maybe it would ease the interface to be able to input list of lists as well.

The template example should be changed as well.

bthirion:
Agreed.

thomasbazeille:
solved by #32

PR #32, opened 2019-09-25
[MRG] API Improvements to solve #22 and #19

thomasbazeille:
#22 is about parallelization back-end. From joblib >=0.12 there is a better way to specify a default parralel back-end for joblib while letting flexibility to user to override it. This PR implement this.

#19 is about input of Template Alignment class. Now it can accept list of lists in the fit() method. (It worked already in the transform). I also added a few input checks for both that raise friendly error messages if needed, and tested those messages


fmralign/template_alignment.py
```diff
@@ -297,21 +292,33 @@ def fit(self, imgs):
...
 
         Returns
         -------
         self
 
         Attributes
         ----------
-        self.template: 4D Niimg object of same shape as one img.
+        self.template: 4D Niimg object
+            Length : n_samples
```
bthirion: what you call Length here is the fourt dimension of the image ?

thomasbazeille: yes, i didn't look up a better name.

Issue #37, opened 2019-12-02
Implementation of reconstruction error ?


emdupre:
Thanks again for making this toolkit !

To understand it a little better, I'm trying to reproduce the Figure 4a from [your recent paper introducing fmralign](https://hal.archives-ouvertes.fr/hal-02278663/); you can see the implementation I have [here](https://github.com/neurodatascience/fmralign-tutorials/blob/master/replicate_bazeille_2019.py), or [launch it remotely with Binder](https://mybinder.org/v2/gh/neurodatascience/fmralign-tutorials/master?filepath=replicate_bazeille_2019.py).

I can't seem to get my distributions for `ridge` or `scaled orthogonal` alignment to match what you have in Figure 4a. I think the most likely explanation is simply that I'm not implementing the formula for reconstruction error appropriately.  

Do you have an implementation of this already available ?

thomasbazeille:
Hello Elizabeth, thanks a lot for your tutorials and the good opportunity to reproduce the results of the paper.  (Sorry in advance for the long message)

I tried to run your code and have several feedbacks:

- the important difference is the experimental set-up. For one pair, we ** use all "ap" conditions to learn one common alignment that we then apply to every "pa" conditions**, whereas you tried to learn one alignment per condition. As a result, ridge fails since the cross-validation scheme requires at least as many images as folds and the results you get with all methods must be spurious

- **The metrics are voxelwise and should be calculated across several samples (conditions here)**. So the results of your implementation of the reconstruction error and ratio match mine at the average level, but importantly, they should be run by voxel and on several samples

- So for plotting you should get 40000 values for each subject. Then the violinplots we did are on all the concatened voxels of target subject. (each point representing one of the 20 x 40000 voxels)
- moreover these metric have no meaning on one sample so they are more reliable calculated on all 

Sidenotes :

- before using the basc atlas to cluster you forgot to resample it so maybe that's what caused you troubles. For my implementation I sticked with automatic kmeans clustering (even though the paper used hierarchical kmeans)

- if you provide a masker to the PairwiseAlignment class with the mask that comes with the data it will avoid fmralign recalculating it.


With these changes I obtain results that are fairly comparable with the paper. (code below)

<img width="506" alt="Capture d’écran 2019-12-13 à 08 47 16" src="https://user-images.githubusercontent.com/6551732/70781447-46456e80-1d86-11ea-8d7f-3f20524ea76a.png">

Full implementation includes:
[...]

PR #40, opened 2020-01-06
[ENH] Adding score metrics

emdupre:
Initial commit of scoring metrics, based on discussion in #37.

To do:
- [x] appropriately handle `sample_weight` in normalized reconstruction
- [x] update documentation accordingly


emdupre:
OK, I think this is ready for review !

Right now I'm still only showcasing correlations in the documentation, but I can update those to use R2 and/or reconstruction error, if you'd like ! Just wasn't sure if it should be in this PR or another one :smile_cat:

fmralign/metrics.py
```diff
@@ -0,0 +1,180 @@
+import numpy as np
+from scipy.stats import pearsonr
+from sklearn.metrics import r2_score
+
+
+def score_table(ground_truth, prediction, masker, loss,
+                multioutput='raw_values'):
```
bthirion: 'multi_output' is more pythonic isn't it ?

fmralign/metrics.py
```diff
@@ -0,0 +1,180 @@
...
+            The R2 distance between source and target arrays.
+            Best possible score is 1.0 and it can be negative (because the
+            model can be arbitrarily worse).
+        'corr' :
+            The correlation between source and target arrays.
+        'n_reconstruction_err' :
+            The normalized reconstruction error. A perfect prediction
+            yields a value of 1.0
+    multioutput: str in ['raw_values', 'uniform_average']
+        Defines aggregating of multiple output scores. Default is raw values.
```
bthirion: applies only if loss is not 'corr'

fmralign/metrics.py
```diff
@@ -0,0 +1,180 @@
...
+    Parameters
+    ----------
+    y_true : arr
+        The ground truth array.
+    y_pred : arr
+        The predicted array.
+    sample_weights : arr
+        Weights to assign to each sample.
+    multioutput: str in ['raw_values', 'uniform_average',
+                         'variance_weighted']
```
bthirion: is variance_weighted used anywhere ?

fmralign/metrics.py
```diff
@@ -0,0 +1,180 @@
...
+                                      denominator[valid_score])
+    if multioutput == 'raw_values':
+        # return scores individually
+        return output_scores
+    
+    elif multioutput == 'uniform_average':
+        # passing None as weights yields uniform average
+        return np.average(output_scores, weights=None)
+
+    elif multioutput == 'variance_weighted':
```
bthirion: Unclear to me what the use case for that is.

fmralign/metrics.py
```diff
@@ -0,0 +1,180 @@
...
+    `Bazeille T., Richard H., Janati H., and Thirion B. (2019) Local
+    Optimal Transport for Functional Brain Template Estimation.
+    In: Chung A., Gee J., Yushkevich P., and Bao S. (eds) Information 
+    Processing in Medical Imaging. Lecture Notes in Computer Science,
+    vol 11492. Springer, Cham.
+    DOI: 10.1007/978-3-030-20351-1_18.`
+    """
+    num = 1 - aligned_error
+    den = 1 - identity_error
+    return 1 - (num / den)
```
bthirion: no need to handle den=0 ?

fmralign/metrics.py
thomasbazeille: multioutput for consistency with api of sklearn r2_score (and metrics in general)

fmralign/metrics.py
thomasbazeille: no apply for all metrics since they are all voxelwise (with time dimension as samples) and the averaging here is across voxels)

fmralign/metrics.py
thomasbazeille: no i'm not sure if it'd be useful but this option makes the API fully compatible with sklearn's r2_score

fmralign/metrics.py
thomasbazeille: see above

thomasbazeille:
I don't know why CircleCI is not running, I'll look into this

fmralign/metrics.py
```diff
@@ -0,0 +1,180 @@
+import numpy as np
+from scipy.stats import pearsonr
+from sklearn.metrics import r2_score
+
+
+def score_table(ground_truth, prediction, masker, loss,
```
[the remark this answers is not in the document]
emdupre: I went with `score_voxelwise` -- let me know what you think !

fmralign/metrics.py
emdupre: Should be handled now :smile_cat: Right now I'm returning 0 if den is 0, but let me know if there's another solution you'd prefer !

emdupre:
Thanks both for your helpful feedback, here ! I think I've integrated most of it. I also kept accidentally adding the wrong files, so I went ahead and added a `.gitignore` to the project. If you'd like that in another PR let me know and I can cherrypick the commit !

I don't have strong opinions on `variance_weighted` -- I agree that it's nice to align with sklearn's interface, but I'm struggling to see a use case for it. I can try it out in the documentation if you'd like ?

fmralign/metrics.py
bthirion: OK, maybe plainly state it that way.

fmralign/metrics.py
bthirion: But if I'm not mistaken this is not used in the code (when loss == 'corr'), is it ?

fmralign/metrics.py
bthirion: OK. It's always a bad  to have some unused / ill-motivated features, but i have no better idea.

fmralign/metrics.py
thomasbazeille: Yes you're right, only the multioutput case is currently handled

fmralign/metrics.py
```diff
@@ -0,0 +1,184 @@
...
+    X_pred = masker.transform(prediction)
+
+    if loss is "R2":
+        score = r2_score(X_gt, X_pred, multioutput=multioutput)
+    elif loss is "n_reconstruction_err":
+        score = normalized_reconstruction_error(
+            X_gt, X_pred, multioutput=multioutput)
+    elif loss is "corr":
+        score = np.array([pearsonr(X_gt[:, vox], X_pred[:, vox])[0]  # pearsonr returns both rho and p
+                          for vox in range(X_pred.shape[1])])
```
thomasbazeille: Related to B.T comment above, multioutput not used

```
if multioutput == "uniform_average":
            score = np.mean(score)
```

should do the job, even though I didn't tested it. It's a pity tests didn't get that, we're not testing uniform_average right ?

emdupre: It doesn't look like CI tests are running at all, currently. Should they be ? pytest is passing locally but I'll add a shape check so the previous commit would have failed !

thomasbazeille:
Sorry for the delay I updated our CircleCI config, pushing an empty commit should run the tests!

emdupre:
> Sorry for the delay I updated our CircleCI config, pushing an empty commit should run the tests!

I pushed something but I don't see anything yet :disappointed: 

Let me know if I can help with the circle config !!

thomasbazeille:
Sorry this time it could work.

emdupre:
Yayyy tests ! :tada: :sparkles: 

I think this is ready for a final review :smile_cat:

.gitignore
```diff
@@ -0,0 +1,127 @@
+# Byte-compiled / optimized / DLL files
```
bthirion: Not a big fan of that (don't have it in my other projects), but no strong feelings.

fmralign/metrics.py
bthirion: I think that the docstring should be a bit mroe rpecise here.

thomasbazeille: it was added in the code since then so now everything is consistent and docstring is right I think

fmralign/metrics.py
```diff
@@ -0,0 +1,166 @@
...
+    Parameters
+    ----------
+    y_true : arr
+        The ground truth array.
+    y_pred : arr
+        The predicted array.
+    sample_weights : arr
+        Weights to assign to each sample.
+    multioutput: str in ['raw_values', 'uniform_average',
+                         'variance_weighted']
```
thomasbazeille:
```suggestion
]
```

fmralign/metrics.py
```diff
@@ -0,0 +1,166 @@
...
+    
+    Parameters
+    ----------
+    y_true : arr
+        The ground truth array.
+    y_pred : arr
+        The predicted array.
+    sample_weights : arr
+        Weights to assign to each sample.
+    multioutput: str in ['raw_values', 'uniform_average',
```
thomasbazeille:
```suggestion
    multioutput: str in ['raw_values', 'uniform_average'
```

fmralign/metrics.py
thomasbazeille:
```suggestion
]
```

emdupre: I committed this outside of the UI so resolving ! :smile_cat:

thomasbazeille:
Any more comments @bthirion  ?

PR #42, opened 2020-01-23
[doc] improve error when calling unrecognized alignment method

emdupre:
Currently, calling an unrecognized alignment method yields `UnboundLocalError`. Since I keep accidentally calling `scaled_procrustes` instead of `scaled_orthogonal`, I thought it might make sense to provide a more interpretable error message !

Also lightly lints for some unused imports.

### DIFF
--- a/fmralign/pairwise_alignment.py
+++ b/fmralign/pairwise_alignment.py
@@ -1,10 +1,7 @@
 # -*- coding: utf-8 -*-
 """ Module for pairwise functional alignment
 """
-
-import copy
 import numpy as np
-from time import time
 
 from sklearn.base import BaseEstimator, TransformerMixin
 from joblib import Parallel, delayed
@@ -101,7 +98,12 @@ def fit_one_piece(X_i, Y_i, alignment_method):
                                        OptimalTransportAlignment,
                                        DiagonalAlignment)):
         alignment_algo = clone(alignment_method)
-    alignment_algo.fit(X_i, Y_i)
+    try:
+        alignment_algo.fit(X_i, Y_i)
+    except UnboundLocalError:
+        warn = "Unrecognized alignment method {}.".format(alignment_method) + \
+               " Please provide a recognized alignment method."
+        raise NotImplementedError(warn)
 
     return alignment_algo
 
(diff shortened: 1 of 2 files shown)


fmralign/pairwise_alignment.py
```diff
@@ -101,7 +98,12 @@ def fit_one_piece(X_i, Y_i, alignment_method):
                                        OptimalTransportAlignment,
                                        DiagonalAlignment)):
         alignment_algo = clone(alignment_method)
-    alignment_algo.fit(X_i, Y_i)
+    try:
+        alignment_algo.fit(X_i, Y_i)
+    except UnboundLocalError:
+        warn = "Unrecognized alignment method {}.".format(alignment_method) + \
+               " Please provide a recognized alignment method."
+        raise NotImplementedError(warn)
```
thomasbazeille: Maybe this would deserve a raise error test?

fmralign/pairwise_alignment.py
```diff
@@ -101,7 +98,12 @@ def fit_one_piece(X_i, Y_i, alignment_method):
                                        OptimalTransportAlignment,
                                        DiagonalAlignment)):
         alignment_algo = clone(alignment_method)
-    alignment_algo.fit(X_i, Y_i)
+    try:
+        alignment_algo.fit(X_i, Y_i)
+    except UnboundLocalError:
+        warn = "Unrecognized alignment method {}.".format(alignment_method) + \
+               " Please provide a recognized alignment method."
```
thomasbazeille: Do we have a list of valid options to suggest ? Otherwise it's great like this

emdupre: I can't find a list... but restructuring so that there is one would make sense ! Might be something for a separate PR, though ? LMK what you think !

fmralign/pairwise_alignment.py
bthirion: +1

fmralign/pairwise_alignment.py
thomasbazeille: I don't think it's a priority at all, thanks a lot for the test!

thomasbazeille:
LGTM merging

PR #43, opened 2020-01-24
[fix] reduce warnings

emdupre:
Right now importing several modules  such as `pairwise_alignment` or `template_alignment` yield `FutureWarning`s. Similarly, running `RidgeCV` alignment yields the following `FutureWarning`:

> /srv/conda/envs/notebook/lib/python3.7/site-packages/sklearn/base.py:434: FutureWarning: The default value of multioutput (not exposed in score method) will change from 'variance_weighted' to 'uniform_average' in 0.23 to keep consistent with 'metrics.r2_score'. To specify the default value manually and avoid the warning, please either call 'metrics.r2_score' directly or make a custom scorer with 'metrics.make_scorer' (the built-in scorer 'r2' uses multioutput='uniform_average').

Unfortunately, this appears on each iteration making outputs harder to find. This PR patches these methods such that future warnings are not raised.

LMK what you think ! :smile_cat:

### DIFF
--- a/fmralign/alignment_methods.py
+++ b/fmralign/alignment_methods.py
@@ -6,6 +6,7 @@ import scipy
 from scipy.spatial.distance import cdist
 from scipy import linalg
 from scipy.sparse import diags
+import sklearn
 from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.utils.linear_assignment_ import linear_assignment
 from sklearn.metrics.pairwise import pairwise_distances
@@ -267,7 +268,9 @@ class RidgeAlignment(Alignment):
             target data
         """
         self.R = RidgeCV(alphas=self.alphas, fit_intercept=True,
-                         normalize=False, scoring=None, cv=self.cv)
+                         normalize=False,
+                         scoring=sklearn.metrics.SCORERS['r2'],
+                         cv=self.cv)
         self.R.fit(X, Y)
         return self
 
--- a/fmralign/alignment_methods.py
+++ b/fmralign/alignment_methods.py
@@ -8,7 +8,7 @@ from scipy import linalg
 from scipy.sparse import diags
 import sklearn
 from sklearn.base import BaseEstimator, TransformerMixin
-from sklearn.utils.linear_assignment_ import linear_assignment
+from scipy.optimize import linear_sum_assignment
 from sklearn.metrics.pairwise import pairwise_distances
 from sklearn.linear_model import RidgeCV
 from joblib import Parallel, delayed
@@ -83,7 +83,8 @@ def optimal_permutation(X, Y):
         transformation matrix
     """
     dist = pairwise_distances(X.T, Y.T)
-    u = linear_assignment(dist)
+    u = linear_sum_assignment(dist)
+    u = np.array(list(zip(*u)))
     permutation = scipy.sparse.csr_matrix(
         (np.ones(X.shape[1]), (u[:, 0], u[:, 1]))).T
     return permutation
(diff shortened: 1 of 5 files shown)


emdupre:
I also realized I was getting an error when importing `_utils` since `packaging` wasn't installed, so I added it as a dependency in https://github.com/Parietal-INRIA/fmralign/pull/43/commits/e1e81e2dc0e4fc9b4eeb8c656450462c30e7cbee !

emdupre:
Also also, a question:

Right now fmralign calls Nibabel's `get_data` in multiple places (i.e. most of the `plot_` examples,  `test_template_alignment`, and in the `fit_one_parcellation` function). This functionality has been deprecated in favor of `get_fdata`. However, I know that Nilearn is retaining an internal copy of the `get_data` functionality to use in their own library.

Do you have a preference on whether we should move to `get_fdata` or Nilearn's `get_data` ?

bthirion:
+1 for 'get_fdata'. 
thx for the rigorous development work !

fmralign/alignment_methods.py
```diff
@@ -6,8 +6,9 @@
 from scipy.spatial.distance import cdist
 from scipy import linalg
 from scipy.sparse import diags
+import sklearn
 from sklearn.base import BaseEstimator, TransformerMixin
-from sklearn.utils.linear_assignment_ import linear_assignment
+from scipy.optimize import linear_sum_assignment
```
bthirion: Does it do exactly the same thing ?

thomasbazeille: yes sklearn people are using it as well in place of their old method and deprecating it in next versions.

thomasbazeille: https://github.com/scikit-learn/scikit-learn/issues/13464

fmralign/alignment_methods.py
```diff
@@ -267,7 +269,9 @@ def fit(self, X, Y):
             target data
         """
         self.R = RidgeCV(alphas=self.alphas, fit_intercept=True,
-                         normalize=False, scoring=None, cv=self.cv)
+                         normalize=False,
+                         scoring=sklearn.metrics.SCORERS['r2'],
```
thomasbazeille: Why did you change the scoring method ? I'm not sure what the default meant.

fmralign/version.py
```diff
@@ -54,6 +54,10 @@
         'min_version': '3.1.1',
         'required_at_installation': True,
         'install_info': _FMRALIGN_INSTALL_MSG}),
+    ('packaging', {
```
thomasbazeille: Great

fmralign/alignment_methods.py
emdupre: It returns a tuple of arrays rather than one array but I [added another line](https://github.com/Parietal-INRIA/fmralign/pull/43/files#diff-130259fe9c0b0a8dd336f7690331449aR87) to update the handling appropriately !

fmralign/alignment_methods.py
emdupre: I changed the scoring method because it was giving a `FutureWarning`

<details>

> /srv/conda/envs/notebook/lib/python3.7/site-packages/sklearn/base.py:434: FutureWarning: The default value of multioutput (not exposed in score method) will change from 'variance_weighted' to 'uniform_average' in 0.23 to keep consistent with 'metrics.r2_score'. To specify the default value manually and avoid the warning, please either call 'metrics.r2_score' directly or make a custom scorer with 'metrics.make_scorer' (the built-in scorer 'r2' uses multioutput='uniform_average').

</details>

Now it should be using the [r2_score](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.r2_score.html) metric directly, so the `multioutput` default is `uniform_average` rather than `variance_weighted` !

Issue #45, opened 2020-02-06
[BUG] If 'clustering' is Nifti and n_pieces !>1, 'clustering' is not used and only one parcel is generated


thomasbazeille:
Length of optimal transport on one example where @emdupre used a pre-computed parcellation was just a symptom of this parcellation not being used and only one big parcel being used.

I will submit a PR to fix that ASAP with test that should capture these kinds of probs.

PR summary :
- will better test 'clustering' type, especially should handle paths as well
- should better test types of clustering inputs
- should better communicate which clustering is used to the user, should raise warning if a big parcel is used and even return an error if optimal transport is tried on a too big parcel to avoid endless computations.

Two other learnings : 
- the clustering is an important part of the pipeline and should be more visible to the user. I diagnosed that with verbose but the prints are not clear. Maybe even with verbose = 0 we should print something that recapitulates the clusters sizes, at least when some parcels are big so that the user is aware of what is going on.
- otherwise I use a test_clustering function so that I can see before hand what parcellations will look like for automatic clustering. I'm thinking of adding this to the package and docs

thomasbazeille:
The PR solving this issue was merged. Closing

PR #46, opened 2020-02-07
[FIX, ENH, MRG] Improve clustering functions, tests, 

thomasbazeille:
- Fix a bug #45 ignoring the nifti clustering when it was given with n_pieces = 1 in PairwiseAlignment
- Refactored everything related to clustering in _make_parcellation function.
- Add more extensive tests for clustering scenarios
- Improve verbose on clustering, raise warning for parcels >1000 voxels and error when trying to align with optimal transport on parcels >5000
- Docstring improvements
- add Elizabeth as a contributor

### DIFF
--- a/fmralign/_utils.py
+++ b/fmralign/_utils.py
@@ -40,12 +40,13 @@ def piecewise_transform(labels, estimators, X):
     X_transform: nd array (n_features, n_samples)
         Transformed data
     """
-
+    unique_labels = np.unique(labels)
     X_transform = np.zeros_like(X)
-    # Labels are from 1 to n where as estimators are indexed from 0 to n-1
-    for i in np.unique(labels):
-        X_transform[:, labels == i] = estimators[i - 1].transform(
-            X[:, labels == i])
+
+    for i in range(len(unique_labels)):
+        label = unique_labels[i]
+        X_transform[:, labels == label] = estimators[i].transform(
+            X[:, labels == label])
     return X_transform
 
 
--- a/fmralign/_utils.py
+++ b/fmralign/_utils.py
@@ -6,7 +6,7 @@ from packaging import version
 from scipy.stats import pearsonr
 from sklearn.cluster import MiniBatchKMeans
 import nilearn
-from nilearn.image import smooth_img, index_img
+from nilearn.image import smooth_img, index_img, new_img_like
 from nilearn.regions.parcellations import Parcellations
 from nilearn.masking import _apply_mask_fmri, intersect_masks
 from nilearn._utils.niimg_conversions import _check_same_fov
@@ -15,7 +15,6 @@ import warnings
 
 def _intersect_clustering_mask(clustering, mask):
     "Take 3D Niimg clustering and bigger mask, output reduced mask"
-    from nilearn.image import new_img_like
     dat = clustering.get_data()
     new_ = np.zeros_like(dat)
     new_[dat > 0] = 1
@@ -183,7 +182,7 @@ def _make_parcellation(imgs, clustering_index, clustering, n_pieces, masker, smo
     elif n_pieces == 1:
         labels = np.ones(
             int(masker.mask_img_.get_fdata().sum()), dtype=np.int8)
-    # otherwise check
+    # otherwise check requested clustering method
     elif clustering == "hierarchical_kmeans" and n_pieces > 1:
         imgs_subset = index_img(imgs, clustering_index)
         if smoothing_fwhm is not None:
@@ -217,7 +216,7 @@ def _make_parcellation(imgs, clustering_index, clustering, n_pieces, masker, smo
     else:
         raise InputError(
             ('Clustering should be "kmeans", "ward", "rena", "hierarchical_kmeans", \
-             or a 3D Niimg and n_pieces an integer ≥ 1'))
+             or a 3D Niimg, and n_pieces should be an integer ≥ 1'))
 
     if verbose > 0:
         unique_labels, counts = np.unique(
--- a/fmralign/_utils.py
+++ b/fmralign/_utils.py
@@ -1,14 +1,16 @@
+# -*- coding: utf-8 -*-
+import os
 import numpy as np
 import nibabel as nib
 from packaging import version
 from scipy.stats import pearsonr
 from sklearn.cluster import MiniBatchKMeans
-
 import nilearn
-from nilearn.image import smooth_img
+from nilearn.image import smooth_img, index_img
 from nilearn.regions.parcellations import Parcellations
 from nilearn.masking import _apply_mask_fmri
 from nilearn._utils.niimg_conversions import _check_same_fov
+import warnings
 
 
 def piecewise_transform(labels, estimators, X):
@@ -42,6 +44,20 @@ def _remove_empty_labels(labels):
     inverse_vals = - np.ones(labels.max() + 1).astype(np.int)
     inverse_vals[vals] = np.arange(len(vals))
     return inverse_vals[labels]
+
+
+def _check_labels(unique_labels, counts, threshold=1000):
+    """ Check is some parcels are bigger than a certain threshold and raise warning if so
+    """
+    above_thr = counts > threshold
+
+    warning = "\n Some parcels are more than 1000 voxels wide it can slow down alignment, especially optimal_transport :"
+    for i in range(len(above_thr)):
+        if above_thr[i]:
+            warning += "\n parcel {} : {} voxels".format(
+                unique_labels[i], counts[i])
+    warnings.warn(warning)
+    pass
 
 
 def _hierarchical_k_means(X, n_clusters, init="k-means++", batch_size=1000,
@@ -110,7 +126,7 @@ def _hierarchical_k_means(X, n_clusters, init="k-means++", batch_size=1000,
     return _remove_empty_labels(fine_labels)
 
 
-def _make_parcellation(imgs, clustering, n_pieces, masker, smoothing_fwhm=5, verbose=0):
+def _make_parcellation(imgs, clustering_index, clustering, n_pieces, masker, smoothing_fwhm=5, verbose=0):
     """Convenience function to use nilearn Parcellation class in our pipeline.
     It is used to find local regions of the brain in which alignment will be later applied.
     For alignment computational efficiency, regions should be of hundreds of voxels.
@@ -119,6 +135,9 @@ def _make_parcellation(imgs, clustering, n_pieces, masker, smoothing_fwhm=5, ver
     ----------
     imgs: Niimgs
         data to cluster
+    clustering_index: list of integers
+        Clustering is performed on a subset of the data chosen randomly
+        in timeframes. This index carry this subset.
     clustering: string or 3D Niimg
         In : {'kmeans', 'ward', 'rena'}, passed to nilearn Parcellations class.
         If you aim for speed, choose k-means (and check kmeans_smoothing_fwhm parameter)
@@ -141,22 +160,30 @@ def _make_parcellation(imgs, clustering, n_pieces, masker, smoothing_fwhm=5, ver
     labels : list of ints (len n_features)
         Parcellation of features in clusters
     """
-    if type(clustering) == nib.nifti1.Nifti1Image:
-        # check image makes suitable labels,
-        # this will return friendly error message if needed
+    # check if clustering is provided
+    if type(clustering) == nib.nifti1.Nifti1Image or os.path.isfile(clustering):
         _check_same_fov(masker.mask_img_, clustering)
-        labels_img = clustering
-    elif clustering == "hierarchical_kmeans":
-        images_to_parcel = imgs
+        labels = _apply_mask_fmri(clustering, masker.mask_img_).astype(int)
+    # otherwise check it's needed, if not return 1 everywhere
+    elif n_pieces == 1:
+        labels = np.ones(
+            int(masker.mask_img_.get_fdata().sum()), dtype=np.int8)
+    # otherwise check
+    elif clustering == "hierarchical_kmeans" and n_pieces > 1:
+        imgs_subset = index_img(imgs, clustering_index)
         if smoothing_fwhm is not None:
-            images_to_parcel = smooth_img(imgs, smoothing_fwhm)
-        X = masker.transform(images_to_parcel)
-        labels_ = _hierarchical_k_means(
+            X = masker.transform(smooth_img(imgs_subset, smoothing_fwhm))
+        else:
+            X = masker.transform(imgs_subset)
+        labels = _hierarchical_k_means(
             X.T, n_clusters=n_pieces, verbose=verbose) + 1
-        labels_img = masker.inverse_transform(labels_)
-    elif clustering in ['kmeans', 'ward', 'rena']:
+
+    elif clustering in ['kmeans', 'ward', 'rena'] and n_pieces > 1:
+        imgs_subset = index_img(imgs, clustering_index)
         if clustering == "kmeans" and smoothing_fwhm is not None:
-            images_to_parcel = smooth_img(imgs, smoothing_fwhm)
+            images_to_parcel = smooth_img(imgs_subset, smoothing_fwhm)
+        else:
+            images_to_parcel = imgs_subset
         try:
             parcellation = Parcellations(method=clustering, n_parcels=n_pieces, mask=masker,
                                          scaling=False, n_iter=20, verbose=verbose)
@@ -168,10 +195,22 @@ def _make_parcellation(imgs, clustering, n_pieces, masker, smoothing_fwhm=5, ver
             else:
                 parcellation = Parcellations(
                     method=clustering, n_parcels=n_pieces, mask=masker, verbose=verbose)
-        parcellation.fit(imgs)
-        labels_img = parcellation.labels_img_
+        parcellation.fit(images_to_parcel)
+        labels = _apply_mask_fmri(
+            parcellation.labels_img_, masker.mask_img_).astype(int)
+
     else:
         raise InputError(
             ('Clustering should be "kmeans", "ward", "rena", "hierarchical_kmeans", \
-             or a 3D Niimg'))
-    return _apply_mask_fmri(labels_img, masker.mask_img_).astype(int)
+             or a 3D Niimg and n_pieces an integer ≥ 1'))
+
+    unique_labels, counts = np.unique(
+        labels, return_counts=True)
+
+    if verbose > 0:
+        print("The alignment will be applied on parcels of sizes {}".format(counts))
+
+    # raise warning if some parcels are bigger than 1000 voxels
+    _check_labels(unique_labels, counts)
+
+    return labels
--- a/fmralign/_utils.py
+++ b/fmralign/_utils.py
@@ -8,9 +8,20 @@ from sklearn.cluster import MiniBatchKMeans
 import nilearn
 from nilearn.image import smooth_img, index_img
 from nilearn.regions.parcellations import Parcellations
-from nilearn.masking import _apply_mask_fmri
+from nilearn.masking import _apply_mask_fmri, intersect_masks
 from nilearn._utils.niimg_conversions import _check_same_fov
 import warnings
+
+
+def _intersect_clustering_mask(clustering, mask):
+    "Take 3D Niimg clustering and bigger mask, output reduced mask"
+    from nilearn.image import new_img_like
+    dat = clustering.get_data()
+    new_ = np.zeros_like(dat)
+    new_[dat > 0] = 1
+    clustering_mask = new_img_like(clustering, new_)
+    return intersect_masks(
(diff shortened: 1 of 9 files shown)


fmralign/_utils.py
```diff
@@ -44,6 +46,20 @@ def _remove_empty_labels(labels):
     return inverse_vals[labels]
 
 
+def _check_labels(unique_labels, counts, threshold=1000):
```
bthirion: maybe simply use `labels` as argument and compute `unique_labels` and `counts` inside ?

fmralign/_utils.py
```diff
@@ -119,6 +135,9 @@ def _make_parcellation(imgs, clustering, n_pieces, masker, smoothing_fwhm=5, ver
     ----------
     imgs: Niimgs
         data to cluster
+    clustering_index: list of integers
+        Clustering is performed on a subset of the data chosen randomly
+        in timeframes. This index carry this subset.
```
bthirion:
```suggestion
        in timeframes. This index carries this subset.
```

fmralign/_utils.py
```diff
@@ -168,10 +195,22 @@ def _make_parcellation(imgs, clustering, n_pieces, masker, smoothing_fwhm=5, ver
...
+            parcellation.labels_img_, masker.mask_img_).astype(int)
+
     else:
         raise InputError(
             ('Clustering should be "kmeans", "ward", "rena", "hierarchical_kmeans", \
-             or a 3D Niimg'))
-    return _apply_mask_fmri(labels_img, masker.mask_img_).astype(int)
+             or a 3D Niimg and n_pieces an integer ≥ 1'))
+
+    unique_labels, counts = np.unique(
```
bthirion: See my comment above

fmralign/alignment_methods.py
```diff
@@ -363,6 +363,9 @@ def fit(self, X, Y):
             target data'''
 
         n = len(X.T)
+        if n > 5000:
```
bthirion: I'm wondering if in that case, one should not spit a warning and return the identity. This would avoid breaking code.
WDYT ?

emdupre:
I won't have time to track this down until the end of the week, but I wanted to ping in that I can't run the [`aly_benchmark`](https://github.com/neurodatascience/fmralign-tutorials/blob/master/aly_benchmark.py) from fmralign-tutorials with this commit. I get an `IndexError` when running the `.transform` method after fitting an alignment.

<details>

```python
/home/emdupre/miniconda3/envs/fmralign/lib/python3.7/site-packages/fmralign/_utils.py:59: UserWarning: 
 Some parcels are more than 1000 voxels wide it can slow down alignment, especially optimal_transport :
  warnings.warn(warning)

---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-6-e74b8709d8c9> in <module>
     10     alignment_estimator.fit(data_dict['source_train'],
     11                             data_dict['target_train'])
---> 12     target_pred = alignment_estimator.transform(data_dict['source_test'])
     13     aligned_score = masker.inverse_transform(score_voxelwise(
     14         data_dict['target_test'], target_pred, masker, loss='corr'))

~/miniconda3/envs/fmralign/lib/python3.7/site-packages/fmralign/pairwise_alignment.py in transform(self, X)
    323         for i in range(self.n_bags):
    324             X_transform += piecewise_transform(
--> 325                 self.labels_[i], self.fit_[i], X_)
    326 
    327         X_transform /= self.n_bags

~/miniconda3/envs/fmralign/lib/python3.7/site-packages/fmralign/_utils.py in piecewise_transform(labels, estimators, X)
     34     # Labels are from 1 to n where as estimators are indexed from 0 to n-1
     35     for i in np.unique(labels):
---> 36         X_transform[:, labels == i] = estimators[i - 1].transform(
     37             X[:, labels == i])
     38     return X_transform

IndexError: list index out of range
```

</details>

I can try to figure out where this is coming from -- and do a full review of this PR -- in a few days if that's helpful !

thomasbazeille:
Thanks a lot I think I know where this is coming from.

Le mardi 11 février 2020, Elizabeth DuPre <notifications@github.com> a
écrit :

fmralign/alignment_methods.py
thomasbazeille: Are warning really enough feedback to say things didn't work as intended at all ? For production it would be a good idea but as long as its experimental it might mislead people testing it on the reason behind unefficiency of the method

bthirion: Well the warning has to be explicit enough.
 
The problem with non-production code is that it never becomes production code.

thomasbazeille:
I solved B.T. comments and bug reported by E.D, ready for review/merge

emdupre:
Thanks, @thomasbazeille ! Running with 9fd33c5 is giving me a `TypeError`:

<details>

```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
TypeError: float() argument must be a string or a number, not 'csr_matrix'

The above exception was the direct cause of the following exception:

ValueError                                Traceback (most recent call last)
<ipython-input-6-74388f44a80f> in <module>
     10     alignment_estimator.fit(data_dict['source_train'],
     11                             data_dict['target_train'])
---> 12     target_pred = alignment_estimator.transform(data_dict['source_test'])
     13     aligned_score = masker.inverse_transform(score_voxelwise(
     14         data_dict['target_test'], target_pred, masker, loss='corr'))

~/Desktop/open_source/fmralign/fmralign/pairwise_alignment.py in transform(self, X)
    337         for i in range(self.n_bags):
    338             X_transform += piecewise_transform(
--> 339                 self.labels_[i], self.fit_[i], X_)
    340 
    341         X_transform /= self.n_bags

~/Desktop/open_source/fmralign/fmralign/_utils.py in piecewise_transform(labels, estimators, X)
     47         label = unique_labels[i]
     48         X_transform[:, labels == label] = estimators[i].transform(
---> 49             X[:, labels == label])
     50     return X_transform
     51 

ValueError: setting an array element with a sequence.
```

</details>

It appears to be sporadic -- with `aly_benchmark` I'm first hitting it at the 18th label (17th estimator). Specifically, running `estimators[17].transform(X[:, labels == 18])` returns:

<details>

```python
array([[<75x75 sparse matrix of type '<class 'numpy.float64'>'
	with 75 stored elements in Compressed Sparse Row format>,
        <75x75 sparse matrix of type '<class 'numpy.float64'>'
	with 75 stored elements in Compressed Sparse Row format>,
        <75x75 sparse matrix of type '<class 'numpy.float64'>'
	with 75 stored elements in Compressed Sparse Row format>,
        ...,
        <75x75 sparse matrix of type '<class 'numpy.float64'>'
	with 75 stored elements in Compressed Sparse Row format>,
        <75x75 sparse matrix of type '<class 'numpy.float64'>'
	with 75 stored elements in Compressed Sparse Row format>,
        <75x75 sparse matrix of type '<class 'numpy.float64'>'
	with 75 stored elements in Compressed Sparse Row format>]], dtype=object)
```

</details>

So, for some reason it's returning sparse instead of dense matrices. Looking into it, currently.

thomasbazeille:
Thanks a lot for the additional testing this is bringing us forward quickly
!

What is the method with which you're testing ? I add one of those problems
with Procrustes sometimes but never managed to reproduce it, since it's
definitely not supposed to return a sparse matrix it must be a peculiar
case (completely empty array or something like this). If you're able to
find a reproducible setting to trigger this behaviour, I'd be glad to give
it a look.

Le ven. 14 févr. 2020 à 23:15, Elizabeth DuPre <notifications@github.com> a
écrit :

emdupre:
> What is the method with which you're testing ? I had one of those problems
with Procrustes but never managed to reproduce it, 

Yes, exactly; it's with Procrustes. Only lightly related: The data I'm testing with in `aly_benchmark` is on OSF, and I've already written a Nilearn-style fetcher that I'm using to access it in `fmralign-tutorials` --  [you can find it here](https://github.com/neurodatascience/fmralign-tutorials/blob/master/fetch_data.py). If it's of interest to add this dataset into the testing suite, I can move over the code and add it into a tutorial !

Here's the particular parcel it fails on, overlaid on the mean functional image for one of the two test subjects:

Screenshot from 2020-02-16 20-46-41

As you can see, this was a very restricted acquisition (the motivating research question focused on the long-axis of the hippocampus), so several of the parcels have no sampled data. So, the calculated alignment returns as a square diagonal matrix that numpy is (smartly ?) keeping compressed. 

The `identity` and `ridge_cv` alignments gives a more meaningful error:

```python
scipy/stats/stats.py:3508: PearsonRConstantInputWarning: An input array is constant; the correlation coefficent is not defined.
  warnings.warn(PearsonRConstantInputWarning())
```

But `optimal_transport` (which now runs ! :horse_racing:) fails silently.

I (personally) think the ideal solution would be to throw a warning to the user that (some of) the specified parcels have constant data over time, and then return zeros for the "calculated" alignment -- or, just drop those parcels. WDYT ?

thomasbazeille:
Ok its not related to this PR but it seems easy to handle.

So after additional testing when X is an array full of 0 :
- `estimator.fit(X, Y)` will never bug (and in most cases learn `R` is identity ).
- `estimator.transform(X)` will work also and return a dense ndarray of zeros for all methods but `ScaledOrthogonalAlignment` (empty CSR) and `DiagonalAlignment` (dense array of nans)
- Then the scoring can throw a warning about constant array but it's a bit unrelated.

So fixing this is easy we should :
- return dense ndarray of zeros for `ScaledOrthogonalAlignment`, `DiagonalAlignment` (e.g if X is empty just learn the alignment as for any other parcel and do nothing if Y is empty as well)

We could also check before every fitting whether a parcel is empty or not. It would be useful to warn or ignore empty parcel alignment(which can be really big if there's a misfit between mask and data and slow down OT a lot). On the other hand if the mask is ill chosen, it's the users responsability and checking every single parcel will add overhead.

What's your opinion @bthirion ?

thomasbazeille:
@emdupre my last commit should solve this,

fmralign/_utils.py
```diff
@@ -141,22 +174,31 @@ def _make_parcellation(imgs, clustering, n_pieces, masker, smoothing_fwhm=5, ver
...
-        labels_img = clustering
-    elif clustering == "hierarchical_kmeans":
-        images_to_parcel = imgs
+        labels = _apply_mask_fmri(clustering, masker.mask_img_).astype(int)
+
+    # otherwise check it's needed, if not return 1 everywhere
+    elif n_pieces == 1:
+        labels = np.ones(
+            int(masker.mask_img_.get_fdata().sum()), dtype=np.int8)
+    # otherwise check
```
emdupre:
```suggestion
    # otherwise check requested clustering method 
```

fmralign/_utils.py
```diff
@@ -168,10 +210,21 @@ def _make_parcellation(imgs, clustering, n_pieces, masker, smoothing_fwhm=5, ver
...
+        parcellation.fit(images_to_parcel)
+        labels = _apply_mask_fmri(
+            parcellation.labels_img_, masker.mask_img_).astype(int)
+
     else:
         raise InputError(
             ('Clustering should be "kmeans", "ward", "rena", "hierarchical_kmeans", \
-             or a 3D Niimg'))
-    return _apply_mask_fmri(labels_img, masker.mask_img_).astype(int)
+             or a 3D Niimg and n_pieces an integer ≥ 1'))
```
emdupre: Light suggestion for clarity:

```suggestion
             or a 3D Niimg, and n_pieces should be an integer ≥ 1'))
```

fmralign/alignment_methods.py
```diff
@@ -363,17 +369,23 @@ def fit(self, X, Y):
             target data'''
 
         n = len(X.T)
-        a = np.ones(n) * 1 / n
-        b = np.ones(n) * 1 / n
+        if n > 5000:
+            warnings.warn(
+                'One parcel is {} voxels. As optimal transport on this region would take too much time, no alignment was performed on it. Decrease parcel size to have intended behavior of alignment.'.format(n))
```
emdupre: Flake8 suggestion:

```suggestion
                'One parcel is {} voxels. As optimal transport on this region '.format(n) +
                'would take too much time, no alignment was performed on it. ' +
                'Decrease parcel size to have intended behavior of alignment.')
```

fmralign/pairwise_alignment.py
```diff
@@ -279,14 +272,25 @@ def fit(self, X, Y):
...
+            # check that clustering provided fills the mask, if not, reduce the mask
+            if 0 in self.masker_.transform(self.clustering):
+                reduced_mask = _intersect_clustering_mask(
+                    self.clustering, self.masker_.mask_img)
+                self.mask = reduced_mask
+                self.masker_ = check_embedded_nifti_masker(self)
+                self.masker_.n_jobs = self.n_jobs
+                self.masker_.fit()
+                warnings.warn(
+                    "Mask used was bigger than clustering provided. Its intersection with the clustering was used instead.")
```
emdupre:
```suggestion
                    "Mask used was bigger than clustering provided. " +
                    "Its intersection with the clustering was used instead.")
```

fmralign/template_alignment.py
```diff
@@ -193,18 +193,19 @@ def __init__(self, alignment_method="identity", n_pieces=1,
...
+            * either 'identity', 'scaled_orthogonal', 'optimal_transport',
+            'ridge_cv', 'permutation', 'diagonal'
+            * or an instance of one of alignment classes (imported from
+            functional_alignment.alignment_methods)
         n_pieces: int, optional (default = 1)
             Number of regions in which the data is parcellated for alignment.
             If 1 the alignment is done on full scale data.
-            If > 1, the voxels are clustered and alignment is performed \
-            on each cluster applied to X and Y.
+            If > 1, the voxels are clustered and alignment is performed on each
```
emdupre: Should we update this docstring to mention the interaction with `clustering` if it's a 3D Niimg ?

fmralign/tests/test_alignment_methods.py
```diff
@@ -151,7 +159,7 @@ def test_all_classes_R_and_pred_shape_and_better_than_identity():
                      ScaledOrthogonalAlignment(scaling=False),
                      OptimalTransportAlignment(),
                      Hungarian(), DiagonalAlignment()]:
-            print(algo)
+            # print(algo)
```
emdupre: Should we set up a logger with a `DEBUG` level for these sorts of statement ? Not for this PR, just thinking about whether it's worth opening an issue.

fmralign/tests/test_pairwise_alignment.py
```diff
@@ -34,6 +43,26 @@ def test_pairwise_identity():
...
+    data_mask = copy.deepcopy(mask_img.get_data())
+    data_mask[0] = 0
+    # create ground truth
+    clustering_mask = new_img_like(mask_img, data_mask)
+    data_clust = copy.deepcopy(data_mask)
+    data_clust[1] = 3
+    # create 2-parcels clustering, smaller than background
+    clustering = new_img_like(mask_img, data_clust)
+
+    # clustering is smaller than mask
```
emdupre: Same question as above -- should we also have a test case where the parcellation is larger than the mask ?

fmralign/tests/test_pairwise_alignment.py
```diff
@@ -34,6 +43,26 @@ def test_pairwise_identity():
...
+    assert (mask_img.get_data() > 0).sum() > (clustering.get_data() > 0).sum()
+    algo = PairwiseAlignment(alignment_method='identity',
+                             mask=mask_img, clustering=clustering)
+    with pytest.warns(UserWarning):
+        algo.fit(img1, img1)
+    assert (algo.mask.get_data() > 0).sum() == (
+        clustering.get_data() > 0).sum()
+    algo.transform(img1)
```
emdupre: It doesn't seem like this transform is being tested ?

fmralign/tests/test_utils.py
```diff
@@ -23,22 +25,46 @@ def test_make_parcellation():
     # make_parcellation is built on Nilearn which already has several test for its Parcellation class
     # here we test just the call of the API is right on a simple example
     img, mask_img = random_niimg((7, 6, 8, 5))
+    indexes = np.arange(img.shape[-1])
     masker = NiftiMasker(mask_img=mask_img).fit()
-    n_pieces = 2
-    # create a predefined parcellation
-    labels_img = nibabel.Nifti1Image(
-        np.hstack([np.ones((7, 3, 8)), 2 * np.ones((7, 3, 8))]), np.eye(4))
+    #
```
emdupre: Extra white space ?

bthirion:
> We could also check before every fitting whether a parcel is empty or not. It would be useful to warn or ignore empty parcel alignment(which can be really big if there's a misfit between mask and data and slow down OT a lot). On the other hand if the mask is ill chosen, it's the users responsability and checking every single parcel will add overhead.

No, I think people will want to rely on pre-defined atlases for all sorts of reasons, and we should certainly raise a warning and return a trivial solution: if the number of parcels is identical in the source and target, the identity is reasonable I think. Otherwise (but this may never be the case), a constant matrix ?

emdupre:
I'm still getting a constant array returned with https://github.com/Parietal-INRIA/fmralign/commit/21a36a7ee784961031de790c1f961b3915a2c704.

I've suggested one potential approach in https://github.com/thomasbazeille/fmralign/pull/1/, where if an empty parcel is found a warning is raised and the `Identity` alignment is returned for that parcel. LMK what you think ! :smile_cat:

fmralign/_utils.py
```diff
@@ -1,14 +1,27 @@
...
+from nilearn.image import smooth_img, index_img
 from nilearn.regions.parcellations import Parcellations
-from nilearn.masking import _apply_mask_fmri
+from nilearn.masking import _apply_mask_fmri, intersect_masks
 from nilearn._utils.niimg_conversions import _check_same_fov
+import warnings
+
+
+def _intersect_clustering_mask(clustering, mask):
+    "Take 3D Niimg clustering and bigger mask, output reduced mask"
```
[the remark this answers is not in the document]
thomasbazeille: Actually it could but in practice it never happens since we mask the clustering image it's already done by default.

fmralign/tests/test_pairwise_alignment.py
thomasbazeille: Actually its already done. Every clustering has at least 0s outside the mask which are removed by masking. But I'm always down for a new test so I will check if this would add something

thomasbazeille:
@emdupre  Yes identity and a constant output is the desired behaviour since you align two arrays full of zeros (parcel out of the brain). But you have no more CSR matrix. Am I right ?

emdupre:
> But you have no more CSR matrix. Am I right ?

Yes, exactly !

But I wasn't getting an identity alignment (looking at the estimators, it was still e.g. a `ScaledOrthogonal` alignment), nor a warning. https://github.com/thomasbazeille/fmralign/pull/1 might be one way to address that, but that might be less relevant for this PR. If so, we could just look at that afterward !

thomasbazeille:
> But I wasn't getting an identity alignment (looking at the estimators, it was still e.g. a ScaledOrthogonal alignment), nor a warning

Yes it's a ScaledOrthogonalAlignment with R (the alignment matrix) which is the identity. I think it's well suited. For the warning yes I didn't include it yet but I was under the impression you proposed it somewhere ?

emdupre:
> I was under the impression you proposed it somewhere ?

Yes, sorry, it's a PR on your fork. GitHub is shortening the link as https://github.com/thomasbazeille/fmralign/pull/1, but the full link is:

```
https://github.com/thomasbazeille/fmralign/pull/1
```

fmralign/_utils.py
```diff
@@ -44,6 +58,22 @@ def _remove_empty_labels(labels):
...
+    """ Check is some parcels are bigger than a certain threshold and raise warning if so
+    """
+    unique_labels, counts = np.unique(
+        labels, return_counts=True)
+
+    if not all(count < threshold for count in counts):
+        warning = "\n Some parcels are more than 1000 voxels wide it can slow down alignment, especially optimal_transport :"
+        for i in range(len(counts)):
+            if counts[i] > threshold:
+                warning += "\n parcel {} : {} voxels".format(
```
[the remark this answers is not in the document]
thomasbazeille: I don't see how to do this. I close for now to not delay the PR but would definitely like to learn a new technique if you have one

fmralign/tests/test_alignment_methods.py
thomasbazeille: maybe not yet, for now the biggest and almost only sources of bugs is clustering labels for various reasons (which is accessible through verbose)

fmralign/tests/test_pairwise_alignment.py
thomasbazeille: I removed it, it's hard to further test

thomasbazeille:
I integrated the warning you proprosed at @emdupre and added a test to assert it's raised when a parcel is 0. Every test is passing, no comment is left, I merge the PR

PR #53, opened 2023-02-16
Updates for newer ott-jax interface

emdupre:
The current codebase uses an older version of OTT-JAX that relies on a deprecated version of JAX ; it is no longer accessible on PyPi and therefore difficult to reproducibly build.

_(NB : I do have a Dockerfile which can be used to this effect ; happy to share here though I think it might be better placed on the fmralign-benchmark repo)._

These changes bump to a more recent version of OTT-JAX and update internal calls accordingly.
I'd like to test a bit more against our older results before committing, but LMK if you have any feedback !

### DIFF
--- a/fmralign/alignment_methods.py
+++ b/fmralign/alignment_methods.py
@@ -4,7 +4,6 @@
 
 import numpy as np
 import scipy
-from scipy.spatial.distance import cdist
 from scipy import linalg
 from scipy.sparse import diags
 import sklearn
@@ -434,14 +433,15 @@ class OptimalTransportAlignment(Alignment):
             '''
         from ott.geometry import geometry
         from ott.solvers.linear import sinkhorn
+        from ott.geometry.costs import Euclidean
         from ott.problems.linear import linear_problem
 
-        cost_matrix = cdist(X.T, Y.T, metric=self.metric)
+        cost_matrix = Euclidean().all_pairs(x=X.T, y=Y.T)
         geom = geometry.Geometry(cost_matrix=cost_matrix, epsilon=self.reg)
-        problem = linear_problem.LinearProblem(geom, X.T, Y.T)
+        problem = linear_problem.LinearProblem(geom)
 
         solver = sinkhorn.Sinkhorn(
-            geom, max_iterations=self.max_iter, threshold=self.tol, jit=True)
+            geom, max_iterations=self.max_iter, threshold=self.tol)
         self.R = solver(problem).matrix
         return self
 
--- a/fmralign/alignment_methods.py
+++ b/fmralign/alignment_methods.py
@@ -402,7 +402,7 @@ class POTAlignment(Alignment):
 
 class OptimalTransportAlignment(Alignment):
     '''Compute the optimal coupling between X and Y with entropic regularization
-    using a OTT as a backend for acceleration.
+    using a OTT-JAX as a backend for acceleration.
 
     Parameters
     ----------
@@ -414,7 +414,7 @@ class OptimalTransportAlignment(Alignment):
 
     Attributes
     ----------
-    R : scipy.sparse.csr_matrix
+    R : jaxlib.xla_extension.Array
         Mixing matrix containing the optimal permutation
     '''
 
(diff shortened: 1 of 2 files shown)


bthirion (approved):
LGTM, thx !

emdupre:
OK, I confirmed that this runs and produces similar results to our previous implementation. Not the same, importantly :

2021 results
```
[0.5555555555555556] | [0.5555555555555556] | [0.42592592592592593] | [0.37037037037037035] | [0.4675925925925926] | [0.36574074074074076]
```

Today's results
```
[0.5462962962962963] | [0.5370370370370371] | [0.4537037037037037] | [0.38425925925925924] | [0.5185185185185185] | [0.39814814814814814]
```

I'm not sure if these are within the range of tolerance we'd expect from a non-fixed seed (which we are not setting here, AFAICT, since I'd expect that we'd have to pass a `PRNGKey` as in other JAX libraries.

bthirion:
The difference is not tiny indeed. Does it correspond to one particular algorithm (sorry I did not follow the technical details) ?

emdupre:
This is staying with a Sinkhorn algorithm but moving from the previous ott-jax version (0.1.6) to a more modern version (I'm testing on 0.4) and applying the associated API changes (they removed the `Transport` interface we were relying on).

I also had to update the cost calculation since it expected a different object that `cdist` was providing -- AFAICT the call I've added is the right one, though other options are [here](https://ott-jax.readthedocs.io/en/latest/_autosummary/ott.geometry.costs.Euclidean.html)

bthirion:
OK, you're playing with OT only. Since it's a convex problem, I would not expect such differences (although I agree they're not dramatic). Maybe the hyperparameter setting has changed too ?

emdupre:
I'll try to list out the internal hyperparam updates -- I'm seeing a pretty significant speed-up, so there are definitely some internal changes.

emdupre:
As a sanity check, I computed the cost_matrix using both the new `Euclidean` interface in ott-jax as well as the older scipy `cdist` call and confirmed that they're passing `np.allclose`, so I don't think that's the cause of the error ; i.e., it really is internal to the updated solver.

emdupre:
I spent a bit of time looking into the parameters, and these are held consistent across versions (though I do think I found a small detail that might partially explain the difference in run times, as the new version is significantly faster ; I've unpacked that a bit below).

I went back to the code itself and stepped through everything again, and I am seeing that the calculated R matrices are exactly the same (with `np.allclose`) between 0.1.6 and 0.4.0. 

So, there's something else that's going on here that's not specific to these ott-jax updates. 

<details>

Updates from ott-jax 0.1.6 (version used in fmralign-benchmark) to 0.4.0 are [visible here](https://github.com/ott-jax/ott/compare/0.1.6...0.4.0) ;  in particular, [this diff](https://github.com/ott-jax/ott/compare/0.1.6...0.4.0#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R68-R73) shows how the sinkhorn example has changed over time.


Based on this [0.1.6 implementation](https://github.com/ott-jax/ott/compare/0.1.6...0.4.0#diff-dacace528e70b19eb1570af021ad003ca1a55cd55fb5a741d00f64b5700d2b04) and the [current implementation](https://github.com/ott-jax/ott/compare/0.1.6...0.4.0#diff-2ff1b80152af4c8d03ff673f22460e5f5e86757eb2d25a7042870539c4943787)  the main thing that jumps out is the `implicit_differentiation` parameter. In the older version [`implicit_differentiation` is True by default](https://github.com/ott-jax/ott/compare/0.1.6...0.4.0#diff-dacace528e70b19eb1570af021ad003ca1a55cd55fb5a741d00f64b5700d2b04L234), whereas in the newer version it's optional and [defaults to unrolling iterations if None](https://github.com/ott-jax/ott/compare/0.1.6...0.4.0#diff-2ff1b80152af4c8d03ff673f22460e5f5e86757eb2d25a7042870539c4943787R702). I'm not sure how much this is accounting for the speed difference between the two.
</details>

bthirion:
Thx for investigating. I would consider that the code is definitely not broken, so let's move ahead  with the change.

emdupre:
@bthirion I saw that you made an update on the JAX calls here ; I've pulled those in and I think this is ready to go ! Let me know if you have any concerns 😸

bthirion:
Yes, the only issue that remains is that CI remains broken, and due tot timeout aI don't know what's happening. But should be good to go indeed.

emdupre:
I want to re-package this, so I'll try to look into that, too 🕵️

PR #61, opened 2023-07-31
[MAINT] Fix tests

emdupre:
This ended up focuses on fixing (relates to #62) and reducing the number of warnings in the test suite. I've also added a GHA to run these on push, so any future failures should be obvious !


bthirion (approved):
LGTM, thx.

emdupre:
I think we need this, so merging !

PR #63, opened 2023-08-08
[MAINT] Use Nilearn's hierarchical kmeans implentation

emdupre:
Since https://github.com/nilearn/nilearn/pull/2282 was merged, we no longer need to ship identical code here in fmralign.

This PR removes the shared implementation in favor of directly calling to Nilearn, as is done with other clustering methods !


emdupre:
Here's the actual error:

> *** ValueError: Found array with 1 feature(s) (shape=(336, 1)) while a minimum of 2 is required by HierarchicalKMeans.


Improve error handling and data shape

emdupre:
The problem was that our test images had only 5 volumes, and we choose a random 20% as the cluster indexing. But, we need at least two for hierarchical_kmeans ! Increasing to 10 volumes to account for that.

emdupre:
All relevant checks are passing, so merging this in now !

PR #64, opened 2023-08-14
[INFRA] Remove Circle in favor of GHA

emdupre:
Now that we have Github Actions up and running, we don't need to rely on the CircleCI infrastructure. I'm removing it here to avoid build redundancy !


emdupre:
I don't think we need to turn anything else off, so merging this !

PR #66, opened 2023-08-16
[INFRA] First pass at doc builder GHA

emdupre:
Try to build docs with GHA and host on gh-pages branch, rather than build locally and host on fmralign-docs.


emdupre:
Doing it live.....

PR #50, opened 2021-11-29
Add Piecewise FastSRM

hugorichard:
This PR is a copy of the piecewise SRM package of Thomas. 
It uses the last version of FastSRM. This is relevant for #49

PR #70, opened 2023-09-05
[ENH] Add fastSRM functionality

emdupre:
Supersedes #50


emdupre:
Opened https://github.com/hugorichard/FastSRM/issues/6 for fastSRM package deprecations !

bthirion:
LMK when you want a review.

emdupre:
@bthirion I think this is now ready ! I've kept the existing PR mostly as-is (just dealing with build and deprecation issues), so would be good to go over it. Will also review myself !

emdupre:
@thomasbazeille I'm guessing you don't have time / interest to review, but your input is also always welcome ! 🤗

fmralign/piecewise.py
```diff
@@ -0,0 +1,457 @@
...
+                piece_srm.add_subjects(X_i, piece_sr)
+        return self
+
+    def transform(self, imgs):
+        """
+        imgs : list of Niimgs or string (paths). Masked shape : n_voxels, n_timeframes
+
+        reshaped_aligned =
+
+        !!!Not implemented for n_bags>1
```
bthirion, line 388: Should we worry about that ? Issue a warning ?

fmralign/piecewise.py
```diff
@@ -0,0 +1,457 @@
...
+        yield [X_[:, i].T for X_ in masked_X_list]
+
+
+def fit_one_piece(piece_X_list, method):
+    """Align source and target data in one piece i, X_i and Y_i, using
+    alignment method and learn transformation to map X to Y.
+
+    Parameters
+    ----------
+    X_i: ndarray
```
bthirion, line 88: The docstring does not march the signature here.

fmralign/piecewise.py
```diff
@@ -0,0 +1,457 @@
...
+
+    if len(reduced_SR) == len(piece_X_list):
+        reduced_SR = np.mean(reduced_SR, axis=0)
+
+    return alignment_algo, reduced_SR
+
+
+def fit_one_parcellation(
+    X_list,
+    srm,
```
bthirion, line 135: alignment_method ?

fmralign/piecewise.py
```diff
@@ -0,0 +1,457 @@
...
+class PiecewiseModel(BaseEstimator, TransformerMixin):
+    """
+    Decompose the source images into regions and summarize subjects information \
+    in a SR, then use alignment to predict \
+    new contrast for target(s) subject.
+    """
+
+    def __init__(
+        self,
+        srm,
```
bthirion, line 203: alignment_method ?

bthirion (commented):
I'm realizing that I'm confused. Are these modules dedicated to SRM specifically or can they be used with other estimators ?

emdupre:
> I'm realizing that I'm confused. Are these modules dedicated to SRM specifically or can they be used with other estimators ?

I believe these can only be used with SRM, since SRM is a group-level method so we have to handle training data differently.

There is already a [fit_one_piece](https://github.com/Parietal-INRIA/fmralign/blob/2af17744f746919c70c45a76d1eb93516a5fc185/fmralign/pairwise_alignment.py#L61) method and (non-overlapping, but similarly named) [generate_Xi_Yi](https://github.com/Parietal-INRIA/fmralign/blob/2af17744f746919c70c45a76d1eb93516a5fc185/fmralign/pairwise_alignment.py#L23) functions in `pairwise_alignment`, which might cause user confusion, though. 

One option is to add another `elif` [here](https://github.com/Parietal-INRIA/fmralign/blob/2af17744f746919c70c45a76d1eb93516a5fc185/fmralign/pairwise_alignment.py#L83-L94), and rename then call out to the `fit_one_piece` function included here.

fmralign/piecewise.py
emdupre: Looks like this was pulled from the existing `fit_one_piece`, which is maybe another plus for integrating it there. Noting that we'll also need to handle `Identity` differently or we'll have overloaded params.

emdupre:
@bthirion I've gone through and patched up the obviously outdated docstrings ; I'm thinking to leave the reorg to another PR, though, just to get the initial code in. WDYT ?

emdupre:
OK, I'm worried that this is now blocking #74 (as we patch dependencies for both nilearn and fastsrm here), so I'd like to get this merged in. Similar to the discussion there, this has SRM in a standalone module.

Please let me know, @bthirion and @denisfouchard, if you have any concerns with merging this ! If I don't hear otherwise, I'll plan to merge early next week.

bthirion:
LGTM. @denisfouchard please voice your opinion.

denisfouchard:
No problems, as SRM is independent from my work. I presume that FastSRM being a more "reputable" package it is ok to keep it standalone. 

How about fugw btw ? Would it be nice to integrate it to compare to basic OT method ? Maybe @alexisthual ?

emdupre:
Thanks, both !

I'm going to go ahead and merge this then, but I opened a new issue with your comment re : fugw so it's not lost ! We can continue the discussion there 🙂

PR #74, opened 2023-12-21
Add Parcel and Searchlight Hyperalignment methods

denisfouchard:
For future reference, this PR is intended to add the Individualized Tuning model ([see paper](https://doi.org/10.1162/imag_a_00032)) into ```fmralign``` and compare it to FastSRM (H. Richard et. al. , 2019)


fmralign/template_alignment.py
```diff
@@ -265,7 +266,7 @@ def __init__(
         alignment_method: string
             Algorithm used to perform alignment between X_i and Y_i :
             * either 'identity', 'scaled_orthogonal', 'optimal_transport',
-            'ridge_cv', 'permutation', 'diagonal'
+            'ridge_cv', 'permutation', 'diagonal', 'pcha', 'slha'
```
bthirion, line 269: I guess that these names are not standard ---- I'm not familiar with  them. Can you provide a more explicit naming ?

bthirion: I guess `searchlight_hyperalignment` and `parcel_hyperalignment`.

bthirion (commented):
Are these things tested or demoed somewhere ?

emdupre:
Hi @denisfouchard,

Thanks for this addition ! Just to +1 @bthirion 's comments, it'd be great to have these new calls explicitly added into new examples and tests. This is adding a new dependency to the library, so it'd be especially useful to showcase why it's a good addition.

Just to note : you'll also need to modify the `pyproject.toml` accordingly ! The test errors are associated with recent nilearn updates, so it's hiding what will later be an import error for the `hyperalignment` package.

In the meantime, I'll update these nilearn calls so your changes are against a passing `main` branch !

denisfouchard:
Hi @emdupre and thanks for the message ! I will review and adapt to your suggestions in the days to come

In the meantime, enjoy your end of the year festivities !

fmralign/template_alignment.py
denisfouchard: yes that is it, i wanted to have short names for different methods

denisfouchard:
Hi @bthirion I will add plots to the package. For the tests, there are in the hyperalignment package I have been working on from the begining ;)

denisfouchard:
There is an other problem, that we will need to adress at some point (maybe at the end of holidays) : we are importing the Hyperalignment package that is going to be private for the rest of its life maybe (as long as Feilong keeps it private at least), so we will need maybe to adapt the code ? We can not just add hyperalignment to the requirements and pip install it directly

bthirion:
We don't want to depend on hyperalignment. We need to copy the relevant code here.

denisfouchard:
> We don't want to depend on hyperalignment. We need to copy the relevant code here.

The problem is that there is a lot of code, and a lot of files. It would imply to copy a ton of files in the package

bthirion:
Yes, but not much more than other estimators. The point is to factorize as much as possible.

denisfouchard:
I'll do my best.
But for example FastSRM is imported, not integrated, and hyperalignment uses a lot of very custom code

bthirion:
FastSRM has been worked out quite well, and is relatively high quality, this is why we can rely on it.

denisfouchard:
I tried to compress and remove all the unnecessary stuff to include it directly in fmralign. Sadly I think I can not go further, as hyperalignment does not seem to fit in the "template_alignment". It is its own unique way of doing alignment

denisfouchard:
If we really want to have the minimum amount of files I can put everything in one file, but this is a bit risky and not very pleasing to read.

fmralign/hyperalignment/model.py
```diff
@@ -0,0 +1,300 @@
...
+            self.regions = parcels
+        else:
+            self.regions = searchlights
+            self.distances = dists
+            self.radius = radius
+
+        # check for cached data
+        try:
+            self.denoised_signal = np.load(
+                self.path + "/train_data_denoised.npy"
```
bthirion, line 130: Same question here. The code feels redundant bzw ?

fmralign/hyperalignment/model.py
```diff
@@ -0,0 +1,300 @@
...
+            numpy.ndarray: The transformed test data.
+        """
+
+        full_signal = np.concatenate(X_test_, axis=1, dtype=np.float32)
+
+        if verbose:
+            print("Predict : Computing stimulus matrix...")
+
+        if self.decomp_method is None:
+            S = stimulus_estimator(
```
bthirion, line 196:
```suggestion
            stimulus_ = stimulus_estimator(
```

fmralign/hyperalignment/model.py
```diff
@@ -0,0 +1,300 @@
...
+
+        if verbose:
+            print("Predict : stimulus matrix shape: ", S.shape)
+
+        reconstructed_signal = [
+            reconstruct_signal(S, T_est) for T_est in self.tuning_data
+        ]
+        return np.array(reconstructed_signal, dtype=np.float32)
+
+    def get_shared_stimulus(self):
```
bthirion, line 208: Sam comment as previously on the getters.

fmralign/hyperalignment/model.py
```diff
@@ -0,0 +1,300 @@
...
+# Computing decomposition
+
+
+def tuning_estimator(shared_response, target, latent_dim=None):
+    """
+    Estimate the tuning weights for individualized neural tuning.
+
+    Parameters:
+    --------
+        - shared_response (array-like):
```
bthirion, line 249: Can you align properly ?

fmralign/hyperalignment/regions_alignment.py
```diff
@@ -0,0 +1,214 @@
...
+            )
+
+            if self.cache:
+                np.save(self.path + (f"/train_data_W_{i}.npy"), x_hat)
+        return x_hat
+
+    def fit_transform(
+        self,
+        X: np.ndarray,
+        regions=None,
```
bthirion, line 107: you should not provide these arguments here.

fmralign/hyperalignment/regions_alignment.py
```diff
@@ -0,0 +1,214 @@
...
+            for i in range(self.n_s)
+        )
+
+        if id is None:
+            id = np.random.randint(0, 1000000)
+
+        if self.cache:
+            np.save(self.path + ("/train_data_denoised.npy"), self.Xhat)
+
+        return np.array(self.Xhat)
```
bthirion, line 200: You don't return the template ?

fmralign/tests/test_alignment_methods.py
```diff
@@ -194,7 +217,67 @@ def test_ott_backend():
...
+    model = INT()
+    model.fit(X_train, searchlights, dists, radius=radius)
+    X_pred = model.transform(X_test)
+
+    # assert that the saved cached files exist
+    b = os.path.exists("cache/")
+    shutil.rmtree("cache")
+    assert b
+    assert X_pred.shape == X_test.shape
+
```
bthirion, line 255: The test is a bit weak. It does not assess whether the code works properly. Can you test the output, not only its shape ?

fmralign/tests/test_alignment_methods.py
```diff
@@ -194,7 +217,67 @@ def test_ott_backend():
...
+    model = INT(n_jobs=-1, alignment_method="parcel")
+    model.fit(X_train, parcels=parcels)
+    X_pred = model.transform(X_test)
+
+    # assert that the saved cached files exist
+    b = os.path.exists("cache/")
+    shutil.rmtree("cache")
+    assert b
+    assert X_pred.shape == X_test.shape
+
```
bthirion, line 279: Same here.

bthirion (commented):
Looks great. I think there is room for improvement, by abiding better to sklearn's conventions.

fmralign/alignment_methods.py
```diff
@@ -465,3 +470,293 @@ def fit(self, X, Y):
...
+    def get_tuning_matrices(self):
+        """
+        Returns the tuning matrices as a NumPy array.
+
+        Returns:
+            numpy.ndarray: The tuning matrices of shape (n_s, latent_dim, n_v) or (n_s, n_t, n_v).
+        """
+        return np.array(self.tuning_data)
+
+    def clean_cache(self, id):
```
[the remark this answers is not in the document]
denisfouchard: i don't think so as it is a custom cache (with numpy matrices)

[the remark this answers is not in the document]
bthirion: Why not replacing that with a clean joblib cache ?

fmralign/alignment_methods.py
```diff
@@ -465,3 +470,293 @@ def fit(self, X, Y):
...
+
+            self.path = path
+
+    def fit(
+        self,
+        X_train,
+        searchlights=None,
+        parcels=None,
+        dists=None,
+        radius: int = 20,
```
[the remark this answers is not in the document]
denisfouchard: not what black tells me interestingly

fmralign/hyperalignment/linalg.py
```diff
@@ -0,0 +1,180 @@
...
+    """
+    U, s, Vt = safe_svd(X, remove_mean=False)
+    d = s / (alpha + s**2)
+    d_UT_Y = d[:, np.newaxis] * (U.T @ Y)
+    betas = Vt.T @ d_UT_Y
+    return betas
+
+
+def procrustes(X, Y, reflection=True, scaling=False):
+    r"""
```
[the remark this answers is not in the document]
denisfouchard: This is Feilong code, and it seems to have special characters, so I can't remove the r sign

fmralign/hyperalignment/regions_alignment.py
denisfouchard: This is the denoising class, so I need to return the denoised signal directly

fmralign/tests/test_alignment_methods.py
denisfouchard: What do you mean ? The test is here just to check that there a no errors on a complete run of the code, and that it returns an output of the type and shape we want, nothing more. What would you want me to add ? (all other tests here do the same thing, we are not doing benchmarks in any of these)

fmralign/alignment_methods.py
```diff
@@ -465,3 +470,293 @@ def fit(self, X, Y):
...
+            self.regions = parcels
+        else:
+            self.regions = searchlights
+            self.distances = dists
+            self.radius = radius
+
+        # check for cached data
+        try:
+            self.denoised_signal = np.load(
+                self.path + "/train_data_denoised.npy"
```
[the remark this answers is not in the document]
denisfouchard: what do you mean consistant ? I mean in the library only simple models are implemented with no need to cache files

bthirion:
LMK when this is ready for review.

denisfouchard:
I think it is. We just have to figure out what is going on with experiments and disappointing results but appart from that code holds on

emdupre:
Sorry, catching up on this conversation !

> : we are importing the Hyperalignment package that is going to be private for the rest of its life maybe (as long as Feilong keeps it private at least), so we will need maybe to adapt the code ?

Just to confirm, then, as I haven't seen that package : are we OK to copy this in re: its licensing ? 

> Sadly I think I can not go further, as hyperalignment does not seem to fit in the "template_alignment". It is its own unique way of doing alignment

I'll look at the code additions to confirm, as "hyperalignment" often means different things to different people. If it's the 2016 Guntupalli implementation, though, I'd say it fits under template alignment, although more iterative in its estimation.

denisfouchard:
Hey Elizabeth ! We have decided not to include the package as a dependency but to add it in a separate folder with the minimum code we need for the experiments. 

It is true that it is by construction a template alignment method, the problem is that the way of computing templates is so different from other methods, that generally use simple tools such as Ridge or Procrustes, that adding Feeling's method in the TemplateAlignment would mean to double the size of the code, and lower the clarity of the code, that is why I think the best solution is to keep a minimal implementation on the side. Let me know what you think !

emdupre:
Thanks, @denisfouchard !

> Hey Elizabeth ! We have decided not to include the package as a dependency but to add it in a separate folder with the minimum code we need for the experiments.

Sorry for being unclear : What I mean is, did you write the code included in this PR in the `hyperalignment` folder, or was it taken from the `hyperalignment` package ? If the latter, we should confirm licensing !

.gitignore
```diff
@@ -1,5 +1,3 @@
-.DS_Store
```
emdupre, line 1: Was there a reason this line was removed ? We generally want to include this in the .gitignore, to make sure it's not accidentally committed by a MacOS user !

bthirion: If I'm not mistaken, the code is actually contributed by us on a branch of the package that is unreleased. I don't think that it is subject to constraints, then ?
The alternative is to recode the functions (there is room for improvement).

denisfouchard:
@emdupre The code I added is mainly mine, as well as the hyperalignment package I was importing before. Basically with @bthirion we took the unachieved and hard to understand codebase and reformated it and reimplemented some parts. So I think no worries with licensing, I cited the paper. @bthirion advised me to remove the name of the original researcher this model was coming from

emdupre:
Got it, thanks !

> If I'm not mistaken, the code is actually contributed by us on a branch of the package that is unreleased. I don't think that it is subject to constraints, then ?
The alternative is to recode the functions (there is room for improvement).

Any code that you've written yourselves is yours to relicense as many times as you see fit, released or unreleased 😸 Anything that you _didn't_ write would be under the `hyperalignment` package license ; which might be compatible with ours, I just wanted to check ! But it sounds like this code is all written by you @bthirion and @denisfouchard, so it should be ok 👍

bthirion:
Please keep it in an ad-hoc module.

bthirion:
LMK when you need another round of reviews.

denisfouchard:
This should be now ready to be merged !

examples/plot_int_alignment.py
```diff
@@ -0,0 +1,256 @@
...
+
+train_index = range(53)
+test_index = range(53, 106)
+full_index = range(106)
+
+train_data = np.array(masked_imgs)[:, train_index, :]
+test_data = np.array(masked_imgs)[:, full_index, :]
+target_test_masked = np.array(masked_imgs)[:, test_index, :]
+
+if False:
```
bthirion, line 143:
```suggestion
use_parcellation = False 
if use_parcellation:
```

examples/plot_int_alignment.py
```diff
@@ -0,0 +1,256 @@
...
+    parcels = compute_parcels(
+        niimg=template_train[0], mask=masker, n_parcels=100, n_jobs=5
+    )
+    model = IndividualizedNeuralTuning(
+        n_jobs=8, alignment_method="parcellation", n_components=None
+    )
+    model.fit(train_data, parcels=parcels, verbose=False)
+    train_stimulus = np.copy(model.shared_response)
+    train_tuning = np.linalg.pinv(train_stimulus) @ model.denoised_signal[-1]
+    model_bis = IndividualizedNeuralTuning(
```
bthirion, line 153: It doe snot make sense to instantiate 2 INTs. By doing so, you're creating two representations without correspondence.

fmralign/alignment_methods.py
```diff
@@ -465,3 +467,239 @@ def fit(self, X, Y):
...
+        n_components=None,
+        alignment_method="searchlight",
+        n_jobs=1,
+    ):
+        """
+        Initialize the IndividualizedNeuralTuning object.
+        Parameters:
+        --------
+
+        - template (str): The type of template to use for alignment. Default is "pca".
```
bthirion, line 493:
```suggestion
        template: str,
                          The type of template to use for alignment. Default is "pca".
```

bthirion: You have to use proper numpydoc conventions

bthirion, line 493: Additionally, you should indicate what are the possible alternatives for this argument.

fmralign/alignment_methods.py
```diff
@@ -465,3 +467,239 @@ def fit(self, X, Y):
...
+            or alignment_method == "ensemble_searchlight"
+        ):
+            self.searchlights = None
+            self.distances = None
+            self.radius = None
+
+        self.tuning_data = []
+        self.denoised_signal = []
+        self.decomp_method = decomp_method
+        self.template = template
```
bthirion, line 524: where is `self.template` used ?

fmralign/alignment_methods.py
```diff
@@ -465,3 +467,239 @@ def fit(self, X, Y):
...
+    # Computing decomposition
+
+    @staticmethod
+    def _tuning_estimator(shared_response, target):
+        """
+        Estimate the tuning weights for individualized neural tuning.
+
+        Parameters:
+        --------
+            - shared_response (array-like):
```
bthirion, line 538: Please use Numpydoc conventions

fmralign/alignment_methods.py
```diff
@@ -465,3 +467,239 @@ def fit(self, X, Y):
...
+        if shared_response.shape[1] < shared_response.shape[0]:
+            return (np.linalg.pinv(shared_response)).dot(target)
+        return np.linalg.inv(shared_response).dot(target)
+
+    @staticmethod
+    def _stimulus_estimator(full_signal, n_t, n_s, latent_dim=None):
+        """
+        Estimates the stimulus response using the given parameters.
+
+        Args:
```
bthirion, line 559: Again

fmralign/alignment_methods.py
```diff
@@ -465,3 +467,239 @@ def fit(self, X, Y):
...
+        U = svd_pca(full_signal)
+        if latent_dim is not None and latent_dim < n_t:
+            U = U[:, :latent_dim]
+
+        stimulus = np.sqrt(n_s) * U
+        stimulus = stimulus.astype(np.float32)
+        return stimulus
+
+    @staticmethod
+    def _reconstruct_signal(shared_response, individual_tuning):
```
bthirion, line 577: This uses nothing from self, so this should be a function (or it is not even needed ?)

fmralign/alignment_methods.py
```diff
@@ -465,3 +467,239 @@ def fit(self, X, Y):
...
+        --------
+            array-like: The estimated tuning weights.
+
+        """
+        if shared_response.shape[1] < shared_response.shape[0]:
+            return (np.linalg.pinv(shared_response)).dot(target)
+        return np.linalg.inv(shared_response).dot(target)
+
+    @staticmethod
+    def _stimulus_estimator(full_signal, n_t, n_s, latent_dim=None):
```
bthirion, line 555: it uses nothing from self, hence  it should be a function ?
Why not use self.n_t, self.n_s instead of giving them as arguments ?

fmralign/alignment_methods.py
```diff
@@ -465,3 +467,239 @@ def fit(self, X, Y):
...
+        - alignment_method (str): The alignment method to use. Can be either "searchlight" or "parcelation", Default is "searchlight".
+        - n_components (int): The number of latent dimensions to use in the shared stimulus information matrix. Default is None.
+        - n_jobs (int): The number of parallel jobs to run. Default is -1.
+
+        Returns:
+        --------
+        None
+        """
+
+        self.n_s = None
```
bthirion, line 504:
```suggestion
        self.n_samples = None
```

fmralign/alignment_methods.py
```diff
@@ -465,3 +467,239 @@ def fit(self, X, Y):
...
+        self.n_t = None
```
bthirion, line 505:
```suggestion
        self.n_time_points = None
```

fmralign/alignment_methods.py
```diff
@@ -465,3 +467,239 @@ def fit(self, X, Y):
...
+        self.n_v = None
```
bthirion, line 506:
```suggestion
        self.n_voxels = None
```

bthirion: Do we need that bzw ?

bthirion: Please make sure you give clear names.

bthirion (commented):
Incomplete pass, but there  is already quite a lot to address.

examples/plot_int_alignment.py
```diff
@@ -0,0 +1,256 @@
...
+        "sub-02",
+        "sub-04",
+        "sub-05",
+        "sub-06",
+        "sub-07",
+    ],
+)
+
+###############################################################################
+# Definine a masker
```
pbarbarant, line 56:
```suggestion
# Define a masker
```
typo

fmralign/alignment_methods.py
```diff
@@ -465,3 +467,239 @@ def fit(self, X, Y):
...
+        self.labels = None
+        self.alphas = None
+        self.alignment_method = alignment_method
+        if alignment_method == "parcelation":
```
pbarbarant, line 510:
```suggestion
        if alignment_method == "parcellation":
```

emdupre:
👋  Hi @denisfouchard, and thanks for your work on this !

Let me know if it's useful for me to review now ; it looks like you and @bthirion are already going back and forth, but I'm happy to provide another set of eyes when you're ready.

fmralign/alignment_methods.py
denisfouchard: s is for subjects here

```suggestion
        self.n_subjects = None
```

fmralign/alignment_methods.py
denisfouchard: You advised me before not to give it self agrs and make it its own function, but for me a static method is easier to access it outside of the model (for experiments for example)

fmralign/alignment_methods.py
bthirion: Sure, but I'm asking you to give the name, because the code is meant to be read by people whio have a hard time understanding the setting. 
For the same reasons, in sklearn, `n_samples` is called `n_samples`, not `n_s`.

denisfouchard:
Hi @emdupre ! Appart from some merging issues that would arise from conflicts in versions of Nilearn calls (the changes you made did not come to my branch logically so I am still using an older version of Nilearn), I think the core code is ready to ship !

denisfouchard:
If it is also good for @bthirion

fmralign/alignment_methods.py
```diff
@@ -13,6 +11,13 @@
 from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.linear_model import RidgeCV
 from sklearn.metrics.pairwise import pairwise_distances
+from .hyperalignment.piecewise_alignment import PiecewiseAlignment
+from .hyperalignment.linalg import safe_svd, svd_pca
+
+import jax
```
emdupre, line 17: We don't want this here ; we only wanted to import jax if it's actually being called, since it's a heavier dependency. That's why it was nested inside the function (and why it's handled separately in the pyproject.toml)

fmralign/alignment_methods.py
```diff
@@ -369,6 +374,7 @@ def fit(self, X, Y):
         Y: (n_samples, n_features) nd array
             target data
         """
+        import ot
```
emdupre, line 377: Why are we moving this inside the function ? It's included in our base dependencies, so it's assumed installed

emdupre (changes requested):
I have not yet made it to the actual INT class, but I wanted to at least start a pass on this ! Thanks again for working on this !!

fmralign/_utils.py
```diff
@@ -71,7 +71,13 @@ def _check_labels(labels, threshold=1000):
 
 
 def _make_parcellation(
-    imgs, clustering_index, clustering, n_pieces, masker, smoothing_fwhm=5, verbose=0
+    imgs,
```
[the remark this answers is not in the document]
denisfouchard: Bertrand told me to black the entire code so that line do not exceed 80 characters. I don't know what to do then...

fmralign/alignment_methods.py
denisfouchard: sorry, just modified that cause the code did not work for me, putting everything back ;)

examples/plot_pairwise_roi_alignment.py
```diff
@@ -54,7 +54,9 @@
 # Select visual cortex, create a mask and resample it to the right resolution
 
 mask_visual = new_img_like(atlas, atlas.get_fdata() == 1)
-resampled_mask_visual = resample_to_img(mask_visual, mask, interpolation="nearest")
+resampled_mask_visual = resample_to_img(
```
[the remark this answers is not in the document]
denisfouchard: Sorry for that, I will try to reformat as previously !

examples/plot_int_alignment.py
```diff
@@ -0,0 +1,256 @@
...
+# Compute a baseline (average of subjects)
+# ----------------------------------------
+# We create an image with as many contrasts as any subject representing for
+# each contrast the average of all train subjects maps.
+#
+
+import numpy as np
+
+masked_imgs = [masker.transform(img) for img in template_train]
+average_img = np.mean(masked_imgs[:-1], axis=0)
```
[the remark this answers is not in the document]
denisfouchard: was a mistake, sorry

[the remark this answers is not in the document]
denisfouchard: there is sub-03 missing, hence the shift in indexes ;)

examples/plot_int_alignment.py
```diff
@@ -0,0 +1,256 @@
...
+# Now we use this scoring function to compare the correlation of predictions
+# made from group average and from template with the real PA contrasts of sub-07
+
+
+average_score = masker.inverse_transform(
+    C_avg := score_voxelwise(target_test, prediction_from_average, masker, loss="corr")
+)
+
+template_score = masker.inverse_transform(
+    C_temp := score_voxelwise(
```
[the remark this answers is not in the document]
denisfouchard: we have to do a masker inverse transform to compute the score voxelwise and also to keep this experiment consistant with the other

examples/plot_int_alignment.py
```diff
@@ -0,0 +1,256 @@
...
+masked_imgs = [masker.transform(img) for img in template_train]
+average_img = np.mean(masked_imgs[:-1], axis=0)
+average_subject = masker.inverse_transform(average_img)
+
+###############################################################################
+# Create a template from the training subjects.
+# ---------------------------------------------
+# We define an estimator using the class TemplateAlignment:
+#   * We align the whole brain through 'multiple' local alignments.
+#   * These alignments are calculated on a parcellation of the brain in 150 pieces,
```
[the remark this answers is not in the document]
denisfouchard: okay, parcellation it will be !

examples/plot_int_alignment.py
```diff
@@ -0,0 +1,256 @@
...
+# ---------------------
+# Finally we plot both scores
+#
+
+# %%
+from nilearn import plotting
+
+baseline_display = plotting.plot_stat_map(
+    average_score, display_mode="z", vmax=1, cut_coords=[-15, -5]
+)
```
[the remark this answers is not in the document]
denisfouchard: Average score is defined (in ```plot_template_alignment.py```) as :

average_score = masker.inverse_transform(
    score_voxelwise(target_test, prediction_from_average, masker, loss="corr")
)

examples/plot_int_alignment.py
denisfouchard: @bthirion This is all the problem with INT : at some point stimulus matrices will loose correspondance. Even if we do co-smoothing with 'closing the square', thus synchronizing stimulus between train and test, there is still no guarantee that the reconstructed image will be in correspondance with the unseen data, as even with a synchronised stimulus, there is still a loss in the order of sources compared to ground truth signal (because we have too cut the stimulus matrix in half at some point to extract the training tuning matrix for the left out subject)

fmralign/hyperalignment/correlation.py
```diff
@@ -0,0 +1,314 @@
...
+
+        else:
+            return (
+                diff_TR_corr.astype(np.float16),
+                same_TR_corr.astype(np.float16),
+                np.array([]),
+                np.array([]),
+            )
+
+    from itertools import combinations
```
emdupre, line 285: Is there a reason this import isn't at the top of the file ?

bthirion:
As you can see, the documentation does not build properly. Can you check what's happening ?
Thx in advance

fmralign/hyperalignment/individualized_neural_tuning.py
```diff
@@ -0,0 +1,150 @@
+from fmralign.alignment_methods import IndividualizedNeuralTuning as BaseINT
+from .regions import compute_searchlights, compute_parcels
+from nilearn.maskers import NiftiMasker
+from nibabel import Nifti1Image
+import numpy as np
+
+
+class IndividualizedNeuralTuning(BaseINT):
```
emdupre, line 8: It looks like we don't have an example using this class ; is that right ? If so, it would be useful to add one just to make it a bit more explicit why this subclass is useful. As is, the existing example uses the BaseINT class, which does work with Niimg objects (as described in the docstring).

Otherwise, I have a light preference to not name the class and the subclass with the same name, but if there's a compelling reason, let me know ? I'd imagine we could call this `HA_INT` or something similar.

fmralign/hyperalignment/local_template.py
```diff
@@ -0,0 +1,262 @@
...
+    """Decompose concatenated data matrices using PCA/SVD.
+
+    Parameters
+    ----------
+    X : ndarray of shape (ns, nt, nv)
+        The input data array.
+    n_components : int or None
+        The number of components to keep. If None, all components are kept.
+    flavor : {'sklearn', 'svd'}
+        Wethter to use sklearn or the custom SVD implementation.
```
emdupre, line 28:
```suggestion
        Whether to use sklearn or the custom SVD implementation.
```

examples/plot_int_alignment.py
```diff
@@ -0,0 +1,219 @@
...
+model = IndividualizedNeuralTuning(
+    n_jobs=8, alignment_method="parcellation", n_components=None
+)
+model.fit(train_data, parcels=parcels, verbose=False)
+train_stimulus = np.copy(model.shared_response)
+train_tuning = np.linalg.pinv(train_stimulus) @ model.denoised_signal[-1]
+model_bis = IndividualizedNeuralTuning(
+    n_jobs=8, alignment_method="parcellation", n_components=None
+)
+model_bis.fit(test_data, parcels=parcels, verbose=False)
```
bthirion, line 154: It still does not make sense to fit test_data by a second model, independently of the training data. I have already written this. Do you understand my comment ?

examples/plot_int_alignment.py
```diff
@@ -0,0 +1,219 @@
...
+test_stimulus = np.copy(model_bis.shared_response)
+# %%
+# We input the mapping image target_train in a list, we could have input more
+# than one subject for which we'd want to predict : [train_1, train_2 ...]
+
+stimulus_ = np.copy(model_bis.shared_response)
+prediction_from_template = stimulus_[test_index] @ train_tuning
```
bthirion, line 161: SHould not be done "manually" by the user.

fmralign/hyperalignment/local_template.py
```diff
@@ -0,0 +1,253 @@
...
+from .linalg import safe_svd
+from .linalg import procrustes
+
+
+def PCA_decomposition(X, max_npc=None, flavor="sklearn", adjust_ns=False, demean=True):
+    """Decompose concatenated data matrices using PCA/SVD.
+
+    Parameters
+    ----------
+    X : ndarray of shape (ns, nt, nv)
```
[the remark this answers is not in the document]
emdupre:
```suggestion
    X : ndarray of shape (n_subjects, n_timesteps, n_voxels)
```

fmralign/hyperalignment/local_template.py
```diff
@@ -0,0 +1,262 @@
+""" Local template computation functions. Those functions are part of the warp hyperalignment
+introducted by Feilong Ma et al. 2023. The functions are adapted from the original code and
+adapted for more general regionations approches.
```
emdupre, line 3: Not clear on the current phrasing, so offering one suggestion here : 

```suggestion
adapted for more general approches.
```

Otherwise could you confirm how this is generalized ?

fmralign/hyperalignment/local_template.py
```diff
@@ -0,0 +1,262 @@
...
+        X_ = X
+    n = min(X_.shape[1], X_.shape[2])
+    n_components = min(n, n_components)
+    XX, cc = PCA_decomposition(
+        X_, n_components=n_components, flavor=flavor, adjust_ns=True, demean=demean
+    )
+    return XX.astype(np.float32)
+
+
+def compute_PCA_var1_template(
```
emdupre, line 120: What is var1 here ? Just a specific variation ?

fmralign/hyperalignment/local_template.py
```diff
@@ -0,0 +1,262 @@
...
+        dss = dss[:, :, sl]
+    XX, cc = PCA_decomposition(
+        dss, n_components=n_components, flavor=flavor, adjust_ns=False, demean=demean
+    )
+    w = np.sqrt(np.sum(cc**2, axis=2)).mean(axis=1)
+    XX *= w[np.newaxis]
+    return XX
+
+
+def compute_procrustes_template(
```
emdupre, line 133: I'd naively think this would go under template alignment. Indeed, `TemplateAlignment` can handle `scaled_orthogonal` alignments. Is there a reason to add this additional function here ?

fmralign/hyperalignment/toy_experiment.py
```diff
@@ -0,0 +1,206 @@
+import os
```
emdupre, line 1: Should we convert this instead to an example ?

fmralign/alignment_methods.py
```diff
@@ -465,3 +469,251 @@ def fit(self, X, Y):
...
+    @staticmethod
+    def _tuning_estimator(shared_response, target):
+        """
+        Estimate the tuning matrix for individualized neural tuning.
+
+        Parameters:
+        --------
+        shared_response : array-like
+             The shared response matrix.
+        target : array-like
```
bthirion, line 539: it would be great if you could specify the matrix dimensions.

emdupre (changes requested):
Thanks again, @denisfouchard ! This is a huge amount of work, so thank you for pushing on it !

I did have a few remaining questions / requests after your follow-up comments (thank you !). My main question is around the HA module. Some of it seems to duplicate functionality of other parts of the code base or of other libraries we already depend on, like scikit-learn. Could we streamline this further ?

Otherwise, HA is not yet included in the doc examples, but there is a `toy_experiment` file that it looks like we could convert pretty easily. Let me know if that sounds reasonable to you ?

fmralign/hyperalignment/toy_experiment.py
bthirion, line 1: What is the status of this module ?

fmralign/tests/test_alignment_methods.py
bthirion: The tests should check that the output is correct (up to numerical variability, but this can be tamed by seeding)

bthirion (commented):
This goes in  the right direction, thx  ! 
But  there is still a significant of work to do on docstrings and examples.

examples/plot_srm_alignment.py
```diff
@@ -0,0 +1,205 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
denisfouchard: We can consider part of this PR ! I did this example to reproduce the same experiment as performed on the INT, as the INT model is not as flexible as SRM.

examples/plot_int_alignment.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+# To infer a template for subjects sub-01 to sub-06 for both AP and PA data,
+# we make a list of 4D niimgs from our list of list of files containing 3D images
+
+from nilearn.image import concat_imgs
+
+template_train = []
+for i in range(6):
+    template_train.append(concat_imgs(imgs[i]))
```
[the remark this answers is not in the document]
denisfouchard: In the INT, we need to train the model also on the target (on the train part) to be able to denoise it. Denoising can only be done on a bunch of subjects, and we need the denoised version of target to compute its tuning matrix

examples/plot_int_alignment.py
```diff
@@ -0,0 +1,256 @@
...
+# -------------------------------------
+# We use target_train data to fit the transform, indicating it corresponds to
+# the contrasts indexed by train_index and predict from this learnt alignment
+# contrasts corresponding to template test_index numbers.
+# For each train subject and for the template, the AP contrasts are sorted from
+# 0, to 53, and then the PA contrasts from 53 to 106.
+#
+
+train_index = range(53)
+test_index = range(53, 106)
```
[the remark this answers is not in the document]
denisfouchard: maybe the name is not appropriate, but in fact we are doing co-smoothing so we need to synchronise sources between sets, that's why we need to compute the test stimulus accross the whole set of contrast. We want to have a basis to extract tuning data from target on one hand, and to reconstruct signal for the target on unseen data. Please let me know if this is still unclear !

fmralign/hyperalignment/correlation.py
```diff
@@ -0,0 +1,314 @@
...
+
+    Returns:
+    ndarray: The similarity matrix between X and Y.
+    """
+    assert X.shape == Y.shape
+    n = X.shape[0]
+
+    def vector_sim(u, v):
+        if metric == "euclidean":
+            return 1 - (np.linalg.norm(u - v) / np.linalg.norm(u + v)) ** 2
```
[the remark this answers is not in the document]
denisfouchard: Sorry for that, anyway those benchmarks are irrelevant now no need to keep them

fmralign/hyperalignment/linalg.py
```diff
@@ -0,0 +1,189 @@
...
+        The solution to the ridge regression problem.
+    """
+    U, s, Vt = safe_svd(X, remove_mean=True)
+    d = s / (alpha + s**2)
+    d_UT_Y = d[:, np.newaxis] * (U.T @ Y)
+    betas = Vt.T @ d_UT_Y
+    return betas
+
+
+def _ridge(X, Y, alpha):
```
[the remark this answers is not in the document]
denisfouchard: we don't, just forgot to delete it

fmralign/hyperalignment/linalg.py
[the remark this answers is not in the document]
denisfouchard: Well this one has de advantage of having all the parameters we need (refleciton and scaling) direclty attached. If I were to use scipy's for example I would have to do a wrapper function to have the same parameter choices. About licencing, I think there is no problem as the actual code is only 4 lines and is a widely known method. Any opinion on that @bthirion ? From my pov, it is the most efficient way to have tailor made code.

fmralign/generate_data.py
```diff
@@ -0,0 +1,153 @@
...
+        S_train = S_std * np.random.randn(n_t, latent_dim)
+        # Normalize each row to have unit norm
+        S_train = S_train / np.linalg.norm(S_train, axis=0, keepdims=True)
+        S_train = S_train @ np.diag(sigma)
+        S_test = S_std * np.random.randn(n_t, latent_dim)
+        S_test = S_test / np.linalg.norm(S_test, axis=0, keepdims=True)
+        S_test = S_test @ np.diag(sigma)
+
+    elif generative_method == "fastsrm":
+        Sigma = rng.dirichlet(np.ones(latent_dim), 1).flatten()
```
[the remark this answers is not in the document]
denisfouchard: What, the distribution ? I just took it from the FastSRM package. Toy data is generated using dirichlet there !

examples/plot_int_alignment.py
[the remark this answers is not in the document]
denisfouchard: I answered in a next comment, please let me know if it is still unclear !

examples/plot_int_alignment.py
denisfouchard: One of the model is only there to do the searchlight hyperalignment denoising, which is indispensable. What I can do is only to import PiecewiseAlignment denoising class.

denisfouchard:
I again tried to address as many remarks as I could, especially doc inconsistencies and int_plot issues. I have still to address the srm_plot, which I will do as soon as I can !

examples/plot_srm_alignment.py
```diff
@@ -0,0 +1,205 @@
...
+# contrasts corresponding to template test_index numbers.
+# For each train subject and for the template, the AP contrasts are sorted from
+# 0, to 53, and then the PA contrasts from 53 to 106.
+#
+test_index = range(53, 106)
+
+test_imgs = np.array(masked_imgs)[:, test_index, :][:-1]
+test_imgs = [x.T for x in test_imgs]
+
+model2 = SRM(n_components=20, n_iter=500)
```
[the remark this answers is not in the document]
denisfouchard: Hmm... they can not be common if we want to predict unseen data from one subject ? For me, temporal compoenents (shared response) must be common but not the basis if we want to avoid leak. That is what I do : I take the shared response from all subjects but the target, while I train the other model on all subjects including the target to get its W_{sub-07} basis.

fmralign/hyperalignment/individualized_neural_tuning.py
denisfouchard: Okay, sounds right ! Indeed we do note have an example with this class as it is just a wrapper, but I don't know which example would be useful. Maybe I shoudl delete it alltogether ?

fmralign/hyperalignment/local_template.py
denisfouchard: Comes from feilong's code but you are right. In fact there is no need for procrustes alignment in the INT model, so I will delete this code

denisfouchard (commented):
Should be all dressed now ! Lmk if something still does not fit

examples/plot_int_alignment.py
```diff
@@ -0,0 +1,212 @@
...
+)
+# %%
+# We input the mapping image target_train in a list, we could have input more
+# than one subject for which we'd want to predict : [train_1, train_2 ...]
+
+
+prediction_from_template = model._reconstruct_signal(
+    shared_response=stimulus_[test_index], individual_tuning=target_tuning
+)
+prediction_from_template = masker.inverse_transform(prediction_from_template)
```
bthirion, line 155:
```suggestion
prediction_from_template_img = masker.inverse_transform(prediction_from_template)
```

fmralign/generate_data.py
```diff
@@ -0,0 +1,153 @@
...
+
+    data_train = np.array(data_train)
+    data_test = np.array(data_test)
+    return data_train, data_test, S_train, S_test, Ts
+
+
+def generate_dummy_searchlights(
+    n_searchlights: int,
+    n_voxels: int,
+    radius: int,
```
bthirion, line 124:
```suggestion
    radius: float,
```

fmralign/hyperalignment/correlation.py
```diff
@@ -0,0 +1,316 @@
...
+        row_ind, col_ind = linear_sum_assignment(corr_mat, maximize=True)
+        corr_mat = corr_mat[row_ind, :]
+        corr_mat = corr_mat[:, col_ind]
+
+    return corr_mat
+
+
+def matrix_MDS(X, Y, n_components=2, dissimilarity="euclidean"):
+    """
+    Perform multidimensional scaling (MDS) on the given data matrices X and Y.
```
bthirion, line 152:
```suggestion
    Perform multidimensional scaling (MDS) on the rows of X and Y.
```

fmralign/hyperalignment/correlation.py
```diff
@@ -0,0 +1,316 @@
...
+
+    X : list of ndarray
+        The first data matrix.
+    Y : list of ndarray
+        The second data matrix.
+    n_components : int
+        The number of dimensions in the output space (default is 2).
+    dissimilarity : str or array-like
+        The dissimilarity measure to use. If it is a string other than "precomputed",
+        the dissimilarity is computed using the Euclidean distance between flattened data points.
```
bthirion, line 165: lines are a bit long

fmralign/hyperalignment/correlation.py
```diff
@@ -0,0 +1,316 @@
...
+    tuple
+        A tuple containing four arrays
+        - corr_same_sub_diff_TR: Correlations between different time points of the same subject.
+        - corr_same_sub_same_TR: Correlations between the same time points of the same subject.
+        - corr_diff_sub_diff_TR: Correlations between different time points of different subjects.
+        - corr_diff_sub_same_TR: Correlations between the same time points of different subjects.
+    """
+    from joblib import Parallel, delayed
+
+    def thread_compute_correlation(X, Y, i, j):
```
bthirion, line 265: Why are you redefining the function ?

fmralign/hyperalignment/local_template.py
```diff
@@ -0,0 +1,206 @@
...
+    flavor : {'sklearn', 'svd'}
+        Whether to use sklearn or the custom SVD implementation.
+    adjust_ns : bool
+        Whether to adjust the variance of the output so that it doesn't increase with the number of subjects.
+    demean : bool
+        Whether to remove the mean of the columns prior to SVD.
+
+    Returns
+    -------
+    XX : ndarray of shape (n_timepoints, n_components)
```
bthirion, line 33: explain what is returned.

fmralign/hyperalignment/test_hyperalignment.py
```diff
@@ -0,0 +1,121 @@
...
+
+    # Check that predicted components have the same shape as original data
+    assert int1.tuning_data[0].shape == (6, int1.n_voxels)
+    assert int2.tuning_data[0].shape == (6, int2.n_voxels)
+    assert int1.shared_response.shape == (int1.n_time_points, 6)
+    assert X_pred.shape == X_test.shape
+
+
+def test_int_with_searchlight():
+    X_train, X_test, stimulus_train, stimulus_test, _ = generate_dummy_signal(
```
bthirion, line 72: Explain what this test does.

fmralign/hyperalignment/toy_experiment.py
```diff
@@ -0,0 +1,203 @@
+"""Toy experiment to test INT on the two parts of the data (ie different runs
```
bthirion, line 1: To be put in the examples.

fmralign/alignment_methods.py
```diff
@@ -465,3 +469,266 @@ def fit(self, X, Y):
...
+        --------
+        ndarray: The reconstructed signal of shape (n_timeframes, n_voxels)
+        (same shape as the original signal)
+        """
+        return (shared_response @ individual_tuning).astype(np.float32)
+
+    def fit(
+        self,
+        X,
+        searchlights=None,
```
bthirion, line 615: All these arguments should be given to the constructor, not to the fit method. Ideally, it only has the X and verbose arguments.

bthirion (commented):
This mostly LGTM.

Can you do a coverage analysis and add tests for parts of the code of this PR that would not yet be tested ?

fmralign/alignment_methods.py
denisfouchard: Well, rewriting the core of the code the last day seems a bit dangerous for me, as a lot of other code depends on it. Will try anyway

bthirion:
Indeed.
I'm assuming that you have enough tests and examples to control what you're doing. 
Anyway, git is your friend, you can always revert changes.

examples/plot_int_alignment.py
```diff
@@ -0,0 +1,212 @@
...
+parcels = compute_parcels(niimg=template_train[0], mask=masker, n_parcels=100, n_jobs=5)
+denoiser = PiecewiseAlignment(alignment_method="parcelation", n_jobs=5)
+denoised_signal = denoiser.fit_transform(X=denoising_data, regions=parcels)
+model = IndividualizedNeuralTuning(
+    n_jobs=8, alignment_method="parcelation", n_components=None
+)
+model.fit(training_data, parcels=parcels, verbose=False)
+stimulus_ = np.copy(model.shared_response)
+target_tuning = model._tuning_estimator(
+    shared_response=stimulus_[train_index], target=denoised_signal[-1]
```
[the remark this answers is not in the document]
denisfouchard: Will explain better. LMK what you think

fmralign/alignment_methods.py
denisfouchard: Edit : done !

examples/plot_int_alignment.py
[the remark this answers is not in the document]
denisfouchard:
```
parcels = compute_parcels(niimg=template_train[0], mask=masker, n_parcels=100, n_jobs=5)
denoiser = PiecewiseAlignment(n_jobs=5)
denoised_signal = denoiser.fit_transform(X=denoising_data, regions=parcels)
target_denoised_data = denoised_signal[-1]
model = IndividualizedNeuralTuning(
    parcels=parcels,
)
model.fit(training_data, verbose=False)
stimulus_ = np.copy(model.shared_response)

# From the denoised data and the stimulus, we can now extract the tuning
# matrix from sub-07 AP contrasts, and use it to predict the PA contrasts.
target_tuning = model._tuning_estimator(
    shared_response=stimulus_[train_index], target=target_denoised_data
)
```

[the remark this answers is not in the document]
bthirion: What troubles me is that you used denoise_signal only for the test subject. This is weird ?

denisfouchard (commented):
.

bthirion (approved):
I think it can be merged now.

denisfouchard:
@emdupre what do u think ? :)

examples/plot_int_alignment.py
```diff
@@ -0,0 +1,216 @@
...
+
+sub_list = ["sub-01", "sub-02", "sub-04", "sub-05", "sub-06", "sub-07"]
+imgs, df, mask_img = fetch_ibc_subjects_contrasts(sub_list)
+
+###############################################################################
+# Define a masker
+# -----------------
+# We define a nilearn masker that will be used to handle relevant data.
+#   For more information, visit :
+#   'http://nilearn.github.io/manipulating_images/masker_objects.html'
```
emdupre, line 52:
```suggestion
#   'https://nilearn.github.io/stable/manipulating_images/masker_objects.html'
```

examples/plot_int_alignment.py
```diff
@@ -0,0 +1,216 @@
...
+
+parcels = compute_parcels(niimg=template_train[0], mask=masker, n_parcels=100, n_jobs=5)
+denoiser = PiecewiseAlignment(n_jobs=5)
+denoised_signal = denoiser.fit_transform(X=denoising_data, regions=parcels)
+target_denoised_data = denoised_signal[-1]
+model = IndividualizedNeuralTuning(
+    parcels=parcels,
+)
+model.fit(training_data, verbose=False)
+stimulus_ = np.copy(model.shared_response)
```
emdupre, line 144: Is there a reason to create this copy ? It doesn't look like we're modifying `model.shared_response` (or `stimulus_`). I'm not sure if `stimulus_` is more commonly used in the INT literature ?

examples/plot_int_alignment.py
```diff
@@ -0,0 +1,216 @@
...
+# Finally we plot both scores
+#
+
+# %%
+from nilearn import plotting
+
+baseline_display = plotting.plot_stat_map(
+    average_score, display_mode="z", vmax=1, cut_coords=[-15, -5]
+)
+baseline_display.title("Group average correlation wt ground truth")
```
emdupre, line 202:
```suggestion
baseline_display.title("Group average correlation wrt ground truth")
```

examples/plot_toy_int_experiment.py
```diff
@@ -0,0 +1,215 @@
...
+
+"""
+Co-smoothing Prediction using the IndividualNeuralTuning Model.
+See article : https://doi.org/10.1162/imag_a_00032
+
+==========================
+
+This is a toy experiment to test Individual Tuning Model (INT) on two parts of the
+data (or different runs) to assess the validity of tuning computation. This code has
+no vocation to be an explanatory example, but rather a test to check the validity of
```
emdupre, line 11:
```suggestion
no intention to be an explanatory example, but rather a test to check the validity of
```

fmralign/alignment_methods.py
```diff
@@ -465,3 +469,265 @@ def fit(self, X, Y):
...
+        n_timepoints = full_signal.shape[0]
+        if scaling:
+            U = svd_pca(full_signal)
+        else:
+            U, _, _ = safe_svd(full_signal)
+        if latent_dim is not None and latent_dim < n_timepoints:
+            U = U[:, :latent_dim]
+
+        stimulus = np.sqrt(n_subjects) * U
+        stimulus = stimulus.astype(np.float32)
```
emdupre, line 616: It looks like this is the first (of 14, by my count) places where we explicitly cast to float32. Is it upcasting to float64 at some point ? Working with nibabel, I'd expect we'd have all float32 arrays to start

fmralign/generate_data.py
```diff
@@ -0,0 +1,153 @@
+"""Some functions to generate toy fMRI data using shared stimulus information."""
```
emdupre, line 1: WDYT of putting these two functions inside `fetch_example_data` instead of in their own file ?

fmralign/hyperalignment/piecewise_alignment.py
```diff
@@ -0,0 +1,157 @@
+"""Piecewise alignment model. This model decomposes the data into regions (pieces).
+Those can either be searchlights or parcels (computed with standard parcellation algorithms).
+See the ```nilearn``` documentation for more details:
+- https://nilearn.github.io/modules/generated/nilearn.regions.Parcellations.html
+- https://nilearn.github.io/dev/modules/generated/nilearn.decoding.SearchLight.html
```
emdupre, line 5:
```suggestion
- https://nilearn.github.io/stable/modules/generated/nilearn.regions.Parcellations.html
- https://nilearn.github.io/stable/modules/generated/nilearn.decoding.SearchLight.html
```

emdupre:
Thank you, @denisfouchard  !! This looks great 🚀  Only small comments, and then I think it's good to go !

examples/plot_int_alignment.py
denisfouchard: In fact *stimulus* is the way to call the shared response in the INT literature !

fmralign/alignment_methods.py
denisfouchard: just to be sure tu have float32 (as we multiply by a potential float64 in sqrt)

emdupre:
> I have made changes according to @emdupre's latest comments. Tbh I don't want to adress now comments about code choices that were made at the begining and that were left uncommented for the whole 3 last months 😅. I did what I could, but I think it is the best we can do for INT !


Thanks, @denisfouchard ! I know that longer reviews are frustrating, but please remember that this is a huge amount of code (for everyone) with very little context (e.g. the conversations on licensing, notification of ready-to-review status, etc) for me.

Regardless, it's fine if you don't want to address at this point any potentially remaining code changes in this PR, but please at least respond to comments brought up in a review so we know if we need to open a dedicated issue.

Thanks again for your work on this.

fmralign/generate_data.py
denisfouchard: Done !

denisfouchard:
> > I have made changes according to @emdupre's latest comments. Tbh I don't want to adress now comments about code choices that were made at the begining and that were left uncommented for the whole 3 last months 😅. I did what I could, but I think it is the best we can do for INT !
> 
> Thanks, @denisfouchard ! I know that longer reviews are frustrating, but please remember that this is a huge amount of code (for everyone) with very little context (e.g. the conversations on licensing, notification of ready-to-review status, etc) for me.
> 
> Regardless, it's fine if you don't want to address at this point any potentially remaining code changes in this PR, but please at least respond to comments brought up in a review so we know if we need to open a dedicated issue.
> 
> Thanks again for your work on this.

Sure, I think I replied to every comment ! (I did move the functions of toy data generations as you proposed)

LMK if it is good for you or if anything should have its own PR if ;)

bthirion:
@emdupre WDYT ?

fmralign/alignment_methods.py
emdupre: I'll open an issue to track, thanks !

emdupre (approved):
I think this is ready 🚀 

Thanks @denisfouchard for the (huge) lift !! Excited to see this added.

bthirion:
Merging, then. Thx to all !

denisfouchard:
Great ! Thank you for the reviews !

PR #76, opened 2024-01-19
[INFRA] Untrack nilearn, set 10.2 as minimum version

emdupre:
We had needed a few unreleased Nilearn updates, but this is no longer the case, so reverts to the safer practice of setting minimum versions.

Here we set 0.10.2 as the minimum, since it's the first version with surface support, which we'd like to integrate soon ! This should update this and get all tests to green.

### DIFF
--- a/fmralign/_utils.py
+++ b/fmralign/_utils.py
@@ -4,9 +4,9 @@ import warnings
 
 import nibabel as nib
 import numpy as np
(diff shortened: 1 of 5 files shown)


emdupre:
Just noting that we still get the sporadic failure in test_template_alignment.... Definitely not yet resolved.

PR #81, opened 2024-02-26
Add SRM cosmoothing experiment

denisfouchard:
As suggested by @bthirion I open this separate PR for an addition of an SRM experiment.

This is co-smoothing on IBC-Contrast (same as INT), inspired by Hugo Richard's code.


denisfouchard:
Sorry for the mess with all the previous commits (feels redundant with the INT PR)

PR #82, opened 2024-02-26
Add SRM experiment

denisfouchard:
As suggested by @bthirion I open this separate PR for an addition of an SRM experiment.

This is co-smoothing on IBC-Contrast (same as INT), inspired by Hugo Richard's code.

### DIFF
--- a/examples/plot_srm_alignment.py
+++ b/examples/plot_srm_alignment.py
@@ -47,7 +47,7 @@ imgs, df, mask_img = fetch_ibc_subjects_contrasts(
 # -----------------
 # We define a nilearn masker that will be used to handle relevant data.
 #   For more information, visit :
-#   'http://nilearn.github.io/manipulating_images/masker_objects.html'
+#   'https://nilearn.github.io/stable/manipulating_images/masker_objects.html'
 #
 
 from nilearn.maskers import NiftiMasker


bthirion (approved):
LGTM, thx.

denisfouchard:
Maybe I can do the merge now ? As it is a single file, independent of the rest and all tests are passing ?

bthirion:
I'd like to give Elizabeth a chance. In general it's good to have two pairs of eyes on the PR.

emdupre:
Sorry for the slow follow-up ; I was sick last week and am still catching up. I'll try to take a look this afternoon

examples/plot_srm_alignment.py
```diff
@@ -0,0 +1,211 @@
...
+#   * The template is created iteratively, aligning all subjects data into a
+#     common space, from which the template is inferred and aligning again to this
+#     new template space.
+#
+
+from nilearn.image import index_img
+from fastsrm.identifiable_srm import IdentifiableFastSRM
+
+
+train_index = range(53)
```
emdupre, line 114: I don't think we've carried over a comment about this being 53 contrasts, so the index seems pretty mysterious to the reader

examples/plot_srm_alignment.py
```diff
@@ -0,0 +1,211 @@
...
+
+# %%
+###############################################################################
+# Predict new data for left-out subject
+# -------------------------------------
+# We use target_train data to fit the transform, indicating it corresponds to
+# the contrasts indexed by train_index and predict from this learnt alignment
+# contrasts corresponding to template test_index numbers.
+# For each train subject and for the template, the AP contrasts are sorted from
+# 0, to 53, and then the PA contrasts from 53 to 106.
```
emdupre, line 140: This is the comment I'd move up, I think !

examples/plot_srm_alignment.py
```diff
@@ -0,0 +1,211 @@
...
+#
+
+
+# To infer a template for subjects sub-01 to sub-06 for both AP and PA data,
+# we make a list of 4D niimgs from our list of list of files containing 3D images
+
+from nilearn.image import concat_imgs
+
+template_train = []
+for i in range(6):
```
emdupre, line 72: Is there a reason why we want to have `sub-07` in `template_train` at all, rather than only in `target_train` and `target_test` ? I know we exclude the last entry later so that it's not actually included in e.g., `average_img`, but that's one extra bit of overhead for the reader to keep in mind.

examples/plot_srm_alignment.py
```diff
@@ -0,0 +1,211 @@
...
+from fastsrm.identifiable_srm import IdentifiableFastSRM
+
+
+train_index = range(53)
+
+
+train_imgs = np.array(masked_imgs)[:, train_index, :]
+train_imgs = [x.T for x in train_imgs]
+
+srm = IdentifiableFastSRM(
```
emdupre, line 120: What's the reasoning for using `IdentifiableFastSRM` over `FastSRM` in this case ?

emdupre (changes requested):
Thank you, @denisfouchard ! This will be great to have !!

I have a few small comments, but my larger question is about how we're handling training and testing data. In [our real-data experiments](https://github.com/neurodatascience/fmralign-benchmark/blob/5e06751a2dfab7d8c0494dabce32a2e0aac31615/fmralignbench/utils.py#L193C1-L206C87), we had used `add_subjects` to add in the `target_train` data after fitting on all of the `sources_train` data. Here, it seems like we're fitting on all training data (including `target_train`) and then extracting the target subject's basis in the fitted `W`. This strikes me as a little harder to follow for the naive reader (since we end up having to ignore the last image in multiple other lines like L95, L144, etc).

Could you confirm if my understanding is correct, and if there's a strong reason to do it this way ?

examples/plot_srm_alignment.py
denisfouchard: Hum, I'd say two main reasons : 
- First, there is no FastSRM estimator class (with sklearn API) to use, and I think it is clearer with an object (for consistancy with other experiments, and easier readability and understanding)
- This is mainly inspired by an experiment written by H. Richard himself
Also, Who can do more can do less ;)

examples/plot_srm_alignment.py
denisfouchard: I don't understand ? We need the data from sub-07 in the training to get its basis, so we need to concat the imgs at some point. Maybe you mean changing the same of this list to simply imgs ? But it would be inconsistant with the INT experiment ?

examples/plot_srm_alignment.py
emdupre: Ah, I see now that the fastSRM package no longer includes a fastSRM class (we had pulled previous code that did, so it's [still preserved in the fmralign-benchmark repo](https://github.com/neurodatascience/fmralign-benchmark/blob/5e06751a2dfab7d8c0494dabce32a2e0aac31615/fmralignbench/fastsrm.py#L1275)). Definitely prefer a sklearn API, so let's keep with this.
Thanks for confirming !

examples/plot_srm_alignment.py
bthirion: What would be your suggestion @emdupre  ?

bthirion: Oh, I see it would be to use `add_subjects`. @denisfouchard can you try and do that ?

emdupre:
I think this should be closed given #162 ; please re-open if this is incorrect !

pbarbarant:
> I think this should be closed given #162 ; please re-open if this is incorrect !

Nope, all good!

PR #85, opened 2024-07-15
Create a deploy action

pbarbarant:
Updates various gh actions and implements automatic deployment of the docs on a gh-pages branch.


bthirion:
DO we need to merge in order to see if it works ?
@emdupre any comment ?

.github/workflows/doc-build.yml
```diff
@@ -82,7 +87,38 @@ jobs:
                 set -o pipefail;
                 make html | tee log.txt;
         -   name: Upload documentation
-            uses: actions/upload-artifact@v3
+            uses: actions/upload-artifact@v4
             with:
-                name: doc
+                name: DocHTML
```
emdupre, line 92: Why this change ?

.github/workflows/doc-build.yml
```diff
@@ -82,7 +87,38 @@ jobs:
...
+        needs: build_docs
+        runs-on: ubuntu-latest
+        steps:
+        -   uses: actions/checkout@v3
+
+        -   name: Download artifacts
+            uses: actions/download-artifact@v4
+            with:
+                name: DocHTML
+                path: doc/_build/html/
```
emdupre, line 107:
```suggestion
                name: DocHTML
                path: doc/_build/html/
                fetch-depth: 0
```

emdupre: Looking at the examples here(https://github.com/ad-m/github-push-action), I think we might be missing `fetch-depth: 0`

.github/workflows/doc-build.yml
pbarbarant: I think it's better to be explicit about the format as sphinx can compile to various forms + it's consistent with other repos in Parietal-INRIA / mind-inria.

.github/workflows/doc-build.yml
pbarbarant: Good point I'm adding it 👍

pbarbarant:
Refactored the `doc-build.yml` with @emdupre to simplify the overall workflow. The website is now deployed from the `gh-pages` branch and will be automatically updated with every merge/push to the `main`branch.
BTW artifacts are also kept.

emdupre:
Thanks, @pbarbarant ! When checks are green I'm good to merge 🚀 

I already set up the redirect from https://parietal-inria.github.io/fmralign-docs and archived that repo, so I think this should be the end of our deployment issues ! 🎉

.github/workflows/doc-build.yml
```diff
@@ -1,95 +1,69 @@
...
-    push:
-        branches:
-        -   main
-    pull_request:
-        branches:
-        -   '*'
+  push:
+    branches:
+      - main
+      - master
```
bthirion, line 7: why master ?

bthirion (commented):
Looks mostly good. Need to wait for CI to complete.

.github/workflows/doc-build.yml
emdupre:
```suggestion
```

.github/workflows/doc-build.yml
```diff
@@ -1,95 +1,69 @@
...
+  pull_request:
+    branches:
+      - '*'
+
+permissions:
+  contents: write
+  pages: write
```
emdupre, line 14: I don't think this is necessary, as the currently running deployment doesn't have it. And as you said, I'd prefer to give minimal permissions if possible !
```suggestion
```

.github/workflows/doc-build.yml
```diff
@@ -1,95 +1,69 @@
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.11
+
+      # Step 3: Install dependencies
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install torch --index-url https://download.pytorch.org/whl/cpu
+          pip install -e ".[doc]"
```
emdupre, line 36: Currently failing as one of the examples calls OTT :

```suggestion
          pip install -e ".[doc, jax, test]"
```

pbarbarant: Good catch!

pbarbarant:
Closes #91

emdupre:
This is now the same error we saw before merging in 7dbab07 , so I think it's ready to go and I'm happy to see this in !

pbarbarant:
Okay I'm merging it!

PR #87, opened 2024-09-26
Add clustering at the beginning of template creation

### LINKED ISSUE #86: Keep a fixed parcellation when computing templates
`_align_images_to_template` instantiates a bunch of `PairwiseAlignment` estimators, each of which generate their own parcellation of the data. I would argue that a single parcellation should be computed once when instantiating the template for the following reasons:

1. Obvious speed gains;
2. The whole dataset could be used to derive more precise regions (especially for ward and kmeans), rather than working with pairs of subjects;
3. Better control when fitting the template.

Best,
PL

pbarbarant:
Closes #86


emdupre:
Do we still want to keep this open given #89 ?

pbarbarant:
> Do we still want to keep this open given #89 ?

No, let's close it since everything will be handled with the new backend.

PR #97, opened 2024-10-30
[DOC] Restrict deploy job to run only on the main branch

pbarbarant:
This is a quickfix to avoid running the deploy action when on a PR.


emdupre:
The failure is a sporadic one we've seen before ; I'll open an issue for it but I think this is OK to merge as-is !

emdupre:
(I've re-launched the checks just to confirm if they fail on a second run-through)

emdupre:
OK, confirmed it was sporadic. Merging !

PR #115, opened 2024-12-03
[FIX] Handle both 3D and 4D images in the case of one contrast

### LINKED ISSUE #114: [BUG] Inconsistence between 3D inputs and 4D templates
When dealing with one contrast, input images are usually 3D however the template is initialized as a 1 sample 4D image which raises an error when comparing the shapes during preprocessing.

pbarbarant:
Closes #114


bthirion (approved):
LGTM, thx

Issue #100, opened 2024-11-03
Add an example showcasing Procrustes barycenter


bthirion:
We've recently observed that Procrustes barycenters (with n > 2 subjects) are quite strong.
I'd like to have an example to showcase that.

Issue #92, opened 2024-10-29
[DOC] Reduce example computation time


emdupre:
Current computation time estimates for all examples :

 | example | time elapsed |
 |--------- | ------------- |
 | plot_template_alignment.py | 1442.31 sec |
 | plot_int_alignment.py | 118.17 sec |
 | plot_alignment_methods_benchmark.py | 85.85 sec |
 | plot_pairwise_alignment.py | 82.16 sec |
 | plot_pairwise_roi_alignment.py | 19.88 sec | 
 | plot_toy_int_experiment.py |  5.61 sec | 
 | plot_alignment_simulated_2D_data.py | 1.79 sec |

We should bring all of these down, but particularly `plot_template_alignment`  !

pbarbarant:
Closed by #116

PR #116, opened 2024-12-08
Refactor template alignments to keep a fixed parcellation

### LINKED ISSUE #86: Keep a fixed parcellation when computing templates

pbarbarant:
Closes #86. Uses the new backend to distribute template computations across parcels.

- [x] Docstrings
- [x] Unit tests
- [x] Adapt `examples/plot_template_alignment.py`

### DIFF
--- a/examples/plot_template_alignment.py
+++ b/examples/plot_template_alignment.py
@@ -1,13 +1,14 @@
 # -*- coding: utf-8 -*-
-
 """
 Template-based prediction.
 ==========================
 
-In this tutorial, we show how to better predict new contrasts for a target
-subject using many source subjects corresponding contrasts. For this purpose,
-we create a template to which we align the target subject, using shared information.
-We then predict new images for the target and compare them to a baseline.
+In this tutorial, we show how to improve inter-subject similarity using a template
+computed across multiple source subjects. For this purpose, we create a template
+using Procrustes alignment (hyperalignment) to which we align the target subject,
+using shared information. We then compare the voxelwise similarity between the
+target subject and the template to the similarity between the target subject and
+the anatomical euclidean average of the source subjects.
 
 We mostly rely on Python common packages and on nilearn to handle
 functional data in a clean fashion.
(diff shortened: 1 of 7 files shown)


pbarbarant:
I refactored the template example to close #100 . Template is now Procrustes-based.

pbarbarant:
The doc build is now around 10min instead of over 30min, which is acceptable if we want to showcase examples with realistic use cases. @emdupre do you think we can close #92 ?

emdupre:
Yes, that seems reasonable to me ! Thank you, @pbarbarant !

examples/plot_template_alignment.py
```diff
@@ -1,13 +1,14 @@
...
-In this tutorial, we show how to better predict new contrasts for a target
-subject using many source subjects corresponding contrasts. For this purpose,
-we create a template to which we align the target subject, using shared information.
-We then predict new images for the target and compare them to a baseline.
+In this tutorial, we show how to improve inter-subject similarity using a template
+computed across multiple source subjects. For this purpose, we create a template
+using Procrustes alignment (hyperalignment) to which we align the target subject,
+using shared information. We then compare the voxelwise similarity between the
+target subject and the template to the similarity between the target subject and
+the anatomical euclidean average of the source subjects.
```
bthirion, line 11:
```suggestion
the anatomical Euclidean average of the source subjects.
```

examples/plot_template_alignment.py
```diff
@@ -64,22 +65,17 @@
 template_train = []
 for i in range(5):
     template_train.append(concat_imgs(imgs[i]))
-target_train = df[df.subject == "sub-07"][df.acquisition == "ap"].path.values
 
-# For subject sub-07, we split it in two folds:
-#   - target train: sub-07 AP contrasts, used to learn alignment to template
-#   - target test: sub-07 PA contrasts, used as a ground truth to score predictions
-# We make a single 4D Niimg from our list of 3D filenames
+# sub-07 will be our left-out subject.
```
bthirion, line 69:
```suggestion
# sub-07 (that is 5th in the list) will be our left-out subject.
```

bthirion (commented):
LGT

PR #123, opened 2025-01-29
[ENH] Replace FUGW by clusterized sparse UOT

pbarbarant:
Replaces FUGW by hard parcellations constraints to build a sparse transport plan. Enables GPU acceleration with torch for fast pairwise/template estimation.

- [x] Docstrings 
- [x] Pairwise alignment
- [x] Template alignment
- [x] CPU tests
- [x] GPU tests

Usage:
```python
from fmralign.sparse_pairwise_alignment import SparsePairwiseAlignment

alignment = SparsePairwiseAlignment(device=device, reg=1e-2, device="cuda")
alignment.fit(img1, img2)
img1_transformed = alignment.transform(img1)
```

### DIFF
--- a/fmralign/_utils.py
+++ b/fmralign/_utils.py
@@ -271,7 +271,7 @@ def _make_parcellation(
     return labels
 
 
-def _create_sparse_cluster_matrix(arr):
+def _sparse_cluster_matrix(arr):
     """
     Creates a sparse matrix where element (i,j) is 1 if arr[i] == arr[j], 0 otherwise.
 
@@ -306,11 +306,13 @@ def _create_sparse_cluster_matrix(arr):
     # Convert to tensors
     rows = torch.tensor(rows)
     cols = torch.tensor(cols)
-    values = torch.ones(len(rows), dtype=torch.float)
+    values = torch.ones(len(rows), dtype=torch.bool)
 
     # Create sparse tensor
     sparse_matrix = torch.sparse_coo_tensor(
-        indices=torch.stack([rows, cols]), values=values, size=(n, n)
+        indices=torch.stack([rows, cols]),
+        values=values,
+        size=(n, n),
     ).coalesce()
 
     return sparse_matrix
--- a/fmralign/_utils.py
+++ b/fmralign/_utils.py
@@ -299,9 +299,8 @@ def _sparse_cluster_matrix(arr):
     # For each value, add all pairs of indices where that value appears
     for indices in value_to_indices.values():
         for i in indices:
-            for j in indices:
-                rows.append(i)
-                cols.append(j)
+            rows += [i] * len(indices)
+            cols += indices
 
     # Convert to tensors
     rows = torch.tensor(rows)
--- a/fmralign/alignment_methods.py
+++ b/fmralign/alignment_methods.py
@@ -10,9 +10,6 @@ import torch
 from fugw.solvers.utils import (
     batch_elementwise_prod_and_sum,
     crow_indices_to_row_indices,
-    solver_ibpp_sparse,
-    solver_mm_l2_sparse,
-    solver_mm_sparse,
     solver_sinkhorn_sparse,
 )
 from fugw.utils import _low_rank_squared_l2, _make_csr_matrix
@@ -783,7 +780,6 @@ class SparseUOT(Alignment):
     def __init__(
         self,
         sparsity_mask,
-        solver="sinkhorn",
         rho=float("inf"),
         reg=0.1,
         max_iter=1000,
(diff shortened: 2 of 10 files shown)


bthirion (commented):
Before going into details about this PR, I'm not 100% sure we should do it.
Isn't it going to make it very hard to compare fugw with alternatives ? What is the core motivation ?

pbarbarant:
> Before going into details about this PR, I'm not 100% sure we should do it. Isn't it going to make it very hard to compare fugw with alternatives ? What is the core motivation ?

As for now, fugw is only included in the alignment methods and thus not compatible with the `Pairwise/TemplateAlignment` API  accessible to the user as it is a whole brain method. One needs to manually process the niftis and extract the volume/mesh to compute the transport cost and run the solver which is not really convenient. I would rather incorporate fugw (and ProMises) in a near future by making a clear distinction with parceled methods.

In the meantime, I can restore the `FUGWAlignment` class in the alignment methods for legacy purposes.

bthirion:
I'm just afraid of losing important functionality. Since we don't have yet any proper comparison btw these methods.
Note that the fact that FUGW works for non-parcelated data is not a major issue: simply require the user to set n_parcels=1 when calling the method.
But if you want to reintroduce it later, that may be OK.

fmralign/tests/test_sparse_template_alignment.py
```diff
@@ -0,0 +1,117 @@
...
+devices = [torch.device("cpu")]
+if torch.cuda.is_available():
+    devices.append(torch.device("cuda:0"))
+
+
+@pytest.mark.skip_if_no_mkl
+@pytest.mark.parametrize(
+    "scale_average, device", product([True, False], devices)
+)
+def test_rescaled_euclidean_mean_torch(scale_average, device):
```
bthirion, line 26: Can you add a 1-line docstring to each test

fmralign/tests/test_sparse_template_alignment.py
```diff
@@ -0,0 +1,117 @@
...
+    assert average_data.dtype == subjects_data[0].dtype
+
+    if scale_average is False:
+        euclidean_mean = torch.mean(torch.stack(subjects_data), dim=0)
+        assert torch.allclose(average_data, euclidean_mean)
+
+
+@pytest.mark.skip_if_no_mkl
+@pytest.mark.parametrize("device", devices)
+def test_align_images_to_template(device):
```
bthirion, line 43: Can you add a 1-line docstring to each test

fmralign/tests/test_utils.py
```diff
@@ -189,3 +194,24 @@ def test_non_contiguous_labels():
     # Test accessing by label
     same_parcel = parceled.get_parcel(1)
     assert_array_equal(same_parcel, expected)
+
+
+def test_sparse_cluster_matrix():
```
bthirion, line 199: Can you add a 1-line docstring to each test

fmralign/alignment_methods.py
```diff
@@ -362,7 +362,7 @@ class POTAlignment(Alignment):
     def __init__(
         self,
         solver="sinkhorn_epsilon_scaling",
-        metric="euclidean",
+        metric="sqeuclidean",
```
bthirion, line 365: This naming comes from POT ?

fmralign/_utils.py
```diff
@@ -267,3 +269,50 @@ def _make_parcellation(
...
+    # Create lists to store indices and values for the sparse matrix
+    rows = []
+    cols = []
+
+    # For each value, add all pairs of indices where that value appears
+    for indices in value_to_indices.values():
+        for i in indices:
+            for j in indices:
+                rows.append(i)
+                cols.append(j)
```
bthirion, line 304: This loop looks very inefficient

fmralign/tests/test_sparse_template_alignment.py
```diff
@@ -0,0 +1,149 @@
...
+        subjects_estimators,
+    )
+    assert len(aligned_data) == len(subjects_data)
+    assert len(subjects_estimators) == len(subjects_data)
+    assert aligned_data[0].shape == subjects_data[0].shape
+
+
+@pytest.mark.skip_if_no_mkl
+@pytest.mark.parametrize("device", devices)
+def test_fit_sparse_template(device):
```
bthirion, line 70: 1-line docstring ?

fmralign/alignment_methods.py
pbarbarant: Scipy actually, but I reverted the change to be consistent with the already existing implementation.

bthirion (approved):
LGTM, thx.

PR #127, opened 2025-02-12
[BACKEND] Remove jax and make POT the default OT implementation

### LINKED ISSUE #95: [ENH] Reduce the number of backends
As for now, we rely on POT and Jax for classical OT and torch for fugw. In order to lighten up the package and ease the maintaining, we should cut out one or more backends.

### LINKED ISSUE #126: [DOC] Add docs on sparse methods + examples
Deploy documentation on the website about `SparsePairwiseAlignment` and `SparseTemplateAlignment`. A comparison with classical ot on an example would be nice also.

pbarbarant:
Closes #95  .

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,12 +28,6 @@ cd fmralign
 pip install -e .
 ```
 
-Note that if you want to use the JAX-accelerated optimal transport methods, you should also run:
-
-```
-pip install fmralign .[jax]
-```
-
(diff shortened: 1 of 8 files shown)


emdupre (approved):
Thank you, @pbarbarant !

bthirion (approved):

PR #130, opened 2025-03-04
[DOC] Add pairwise surface alignment example

pbarbarant:
This PR adds an example of Procrustes alignment on the fsaverage5 surface using Ward clustering. An interactive matplotlib animation is added at the end of the notebook to progressively watch the individual idiosyncrasies get smoothed out by the alignment process.

### DIFF
--- a/examples/plot_surf_alignment.py
+++ b/examples/plot_surf_alignment.py
@@ -1,18 +1,16 @@
 # -*- coding: utf-8 -*-
-# %%
 """
-Pairwise functional alignment.
+Pairwise surface alignment.
 ==============================
 
-In this tutorial, we show how to better predict new contrasts for a target
-subject using source subject corresponding contrasts and data in common.
+In this tutorial, we show how to align surface data from two subjects using
+a pairwise alignment method. We project the data on the `fsaverage5` surface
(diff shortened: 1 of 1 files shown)


examples/plot_surf_alignment.py
```diff
@@ -0,0 +1,205 @@
...
+
+###############################################################################
+# Fitting the alignment operator
+# ------------------------------
+# We use the `PairwiseAlignment` class to learn the alignment operator from
+# one subject to the other. We select the `scaled_orthogonal` method to compute
+# a rigid piecewise alignment mapping and the `ward` clustering method to
+# parcellate the cortical surface.
+
+from fmralign.pairwise_alignment import PairwiseAlignment
```
bthirion, line 71: When I run this I get a 
```
ImportError: cannot import name 'concatenate_surface_images' from 'nilearn.maskers._utils'
```

bthirion (commented):
The Nilearn dependecy is not perfectly tracked.

examples/plot_surf_alignment.py
pbarbarant: Which nilearn version are you using? I'm on `0.11.1` which is the last release.

PR #135, opened 2025-05-16
[ENH] Add saving and loading utilities for alignment estimators

pbarbarant:
This short PR provides two new utilities for saving and loading fitted estimators. Usage is as such:

```python
from fmralign._utils import save_alignment, load_alignment

estimator = PairwiseAlignment().fit(X, Y)
save_alignment(estimator, "save_dir")
loaded_estimator = load_alignment("save_dir")
```


bthirion (approved):
The PR LGTM. I am wondering whether we want to document the provenance of the models.

pbarbarant:
> The PR LGTM. I am wondering whether we want to document the provenance of the models.

The user has the option to input a custom name, otherwise the model name is given a timestamp. Once loaded, one can print all the attributes (`n_pieces, alignment_method`, etc...) by accessing the class. In my opinion, documenting the various parameters would result in long names or over-nested folders.

PR #144, opened 2025-07-02
[MAINT, FIX] Move nilearn pin

emdupre:
Resolves test failures with latest release of nilearn

Extra changes to pyproject.toml introduced when touching file, given new pre-commit rules (pulled to match nilearn's rules).

### DIFF
--- a/fmralign/_utils.py
+++ b/fmralign/_utils.py
@@ -1,7 +1,10 @@
 # -*- coding: utf-8 -*-
+import datetime
 import warnings
 from collections import defaultdict
+from pathlib import Path
 
+import joblib
 import nibabel as nib
 import numpy as np
 import torch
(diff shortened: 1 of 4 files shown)


emdupre:
Currently our tests are failing because of a Nilearn deprecation :
>[:bdg-danger:`Deprecation`](https://github.com/nilearn/nilearn/blob/main/doc/changes/0.12.0.rst#id312) Using the transform method of nifti maskers on 3D images now returns 1D arrays. Similarly, using the transform method of surface maskers on 1D surface images now returns 1D arrays. ([:gh:`5381`](https://github.com/nilearn/nilearn/blob/main/doc/changes/0.12.0.rst#id314) by [`Rémi Gau`_](https://github.com/nilearn/nilearn/blob/main/doc/changes/0.12.0.rst#id448)).

In [6e02238](https://github.com/fmralign/fmralign/pull/144/commits/6e022389a81e8b5c8da62548142dd0713440a91a) I've addressed the majority of our use cases, but the remaining failure is that we're now getting 1D tensors when FUGW (called in sparseUOT) expects 2D tensors.

We might want to ensure FUGW is compatible with the newest nilearn release:  If yes, pad our tensors here ; if no, update FUGW.

emdupre:
@pbarbarant I don't have a strong handling on the sparse alignment methods, so I've done a pretty minimal patch here. There's certainly something more elegant we could do, if you have suggestions !

pbarbarant (approved):
LGTM !

PR #145, opened 2025-07-04
Remove Identity, Ridge from benchmark, move to Template

ferponcem:
(My Hello World for fmralign 🎉)

First changes to the benchmark example:

- Moving from *PairwiseAlignment* to *TemplateAlignment*
- Keeping only `scaled_orthogonal` and `optimal_transport` as methods
- Minimal changes to the texts 
- (extra) Removed the colorbar from the ROI plot, since it doesn't render correctly (to be discussed)

**Note:** The new nilearn's default cmap has changed to `cmap='RdBu_r'`, let me know if you'd prefer me to go back to the cmap of previous examples.

**Working on:** Adding *FastSRM* to this example


emdupre:
Thank you, @ferponcem !

The current test errors were resolved in #144 ; could you please pull from `main` to add in those changes, so we can see the rendered examples ?

emdupre:
The doc build has now been running for over 3h ; it looks like the majority of this time has been on `plot_pairwise_roi_alignment.py`. We haven't seen this on other PRs, so will see if re-running resolves or if this is indeed a real issue.

pbarbarant:
> Hey ! I just pushed some lines to include SRM to this example, but I am aware that the doc build is not working. I am unsure if I am able to interrupt if myself, I hope you can do it while we try to see why that's the case? @emdupre @pbarbarant

Hi @ferponcem , we discovered that the default parcellation method was kmeans instead of ward which greatly slowed down `TemplateAlignment` with Optimal Transport (because some parcels were > 1000 voxels) , #155 should have fixed it.

ferponcem:
Hey ! I have just tried increasing the number of training subjects to 2 for all methods, and testing on a third one.
I've done so to decrease computation time. For the same purpose I've increased `n_pieces`.
I tried to make it very clear but maybe it's too much text.
And for having the 3 figures together, I am plotting them as subplots. But of course this can be done differently.
Please let me know what you think 🚀 
And thanks for all the suggestions !!

examples/plot_alignment_methods_benchmark.py
```diff
@@ -163,11 +145,109 @@
...
+# Initialize the IdentifiableFastSRM model. This version of SRM ensures that
+# the solution is unique.
+#
+
+from fastsrm.identifiable_srm import IdentifiableFastSRM
+
+srm = IdentifiableFastSRM(
+    n_components=30,
+    n_iter=10,
+)
```
emdupre, line 168: We can move this up to L126, so we define all the methods in the same section !

examples/plot_alignment_methods_benchmark.py
```diff
@@ -163,11 +145,109 @@
...
+
+source_subjects = [sub for sub in sub_ids if sub != "sub-04"]
+source_train = [
+    concat_imgs(df[(df.subject == sub) & (df.acquisition == "ap")].path.values)
+    for sub in source_subjects
+]
+target_train = concat_imgs(
+    df[(df.subject == "sub-04") & (df.acquisition == "ap")].path.values
+)
+
```
emdupre, line 191: This can be removed since it now overlaps with the subject definition for the other methods !

examples/plot_alignment_methods_benchmark.py
```diff
@@ -163,11 +145,109 @@
...
+#
+
+source_test = [
+    concat_imgs(df[(df.subject == sub) & (df.acquisition == "pa")].path.values)
+    for sub in source_subjects
+]
+target_test = concat_imgs(
+    df[(df.subject == "sub-04") & (df.acquisition == "pa")].path.values
+)
+
```
emdupre, line 207: Similarly, this can also be removed in favor of L90 !

examples/plot_alignment_methods_benchmark.py
```diff
@@ -75,35 +75,19 @@
...
-# * source train: AP contrasts for subject sub-01
-# * target train: AP contrasts for subject sub-02
-#
+# Let's organize our data into training and testing sets:
+# * Training set: Files from the first two subjects (sub-01, sub-02) with both
+#   AP and PA acquisitions. These will serve as our source subjects for
+#   alignment.
+# * Testing set: Files from the third subject (sub-04) with both AP and PA
+#   acquisitions, which will be our target subject for evaluating alignment
+#   quality.
```
emdupre, line 86: Note that we actually do want to keep a separate split between AP and PA acquisitions so we can have a "train" and a "test" data set for both our "source" subjects and "target" subjects ! So we can re-use the original PA / AP split, just now with one additional subject.

examples/plot_alignment_methods_benchmark.py
```diff
@@ -163,11 +145,109 @@
...
+aligned_test = srm.transform([roi_masker.transform(target_test).T])
+aligned_pred = roi_masker.inverse_transform(
+    srm.inverse_transform(aligned_test[0])[0].T
+)
+
+# Step 4: Evaluate voxelwise correlation between predicted and true test
+# signals.
+srm_error = score_voxelwise(
+    target_test, aligned_pred, masker=roi_masker, loss="corr"
+)
```
emdupre, line 232: This (and L233) can be moved around L132

examples/plot_alignment_methods_benchmark.py
```diff
@@ -163,11 +145,109 @@
...
+    display_mode="z",
+    cut_coords=[-15, -5],
+    vmax=1,
+    title=title,
+    axes=axes[len(methods)],
+    colorbar=True,
+)
+
+plt.tight_layout()
+plt.show()
```
emdupre, line 247: We won't need this additional plotting, since we'll include it in the first plot as a sub-plot

examples/plot_alignment_methods_benchmark.py
```diff
@@ -163,11 +145,109 @@
         cut_coords=[-15, -5],
         vmax=1,
         title=title,
+        axes=axes[i],
+        colorbar=False,
     )
```
emdupre, line 150: We can insert the final SRM plot here, just outside the for loop:

```
title = "Correlation of prediction after SRM alignment"
display = plotting.plot_stat_map(
    srm_score,
    display_mode="z",
    cut_coords=[-15, -5],
    vmax=1,
    title=title,
    axes=axes[-1],
    colorbar=False,
)
```

examples/plot_alignment_methods_benchmark.py
ferponcem: In L90 we're not splitting the "ap" and "pa" acquisitions... I can modify this lines so the name matches? or I can do the split since the beginning and just concatenate all of them for the template method

emdupre: Thanks for confirming ! 

I definitely see the confusion, and you're right that for most TemplateAlignment methods we could combine the AP-PA acquisitions, but for fairness I think that have to leave out `source_test` (otherwise two of the methods will learn on more data, so they should innately do better than SRM). It would just make the results harder to compare !

ferponcem:
Just saw it's still not handling the plotting properly, working on it !

ferponcem:
Okay ! I think I am finally getting somewhere.
I struggled to make the subplot work, dealing with axes and delaying `plot_stat_map` is not so straightforward.
So, instead I am trying the not-so-elegant solution of keeping the results and only plotting at the end.
Let me know what you think !

bthirion:
This is starting to look good. Please described a bit mroe the data: which are the features (contrasts) used ?

bthirion:
Indeed, you want to keep it concise. Just give the reader a rough idea of what is there.

examples/plot_alignment_methods_benchmark.py
```diff
@@ -125,29 +138,39 @@
 # ---------------------------------------------------
 # On each region, we search for a transformation R that is either :
 #   *  orthogonal, i.e. R orthogonal, scaling sc s.t. ||sc RX - Y ||^2 is minimized
-#   *  a ridge regression : ||XR - Y||^2 + alpha *||R||^2 with a L2 penalization
-#      on the norm of R.
 #   *  the optimal transport plan, which yields the minimal transport cost
 #       while respecting the mass conservation constraints. Calculated with
 #       entropic regularization.
```
emdupre, line 143: Somehow the formatting on this one had gotten offset :

```suggestion
#   *  the optimal transport plan, which yields the minimal transport cost
#       while respecting the mass conservation constraints. Calculated with
#       entropic regularization.
```

examples/plot_alignment_methods_benchmark.py
```diff
@@ -156,18 +179,61 @@
...
+# Step 4: Evaluate voxelwise correlation between predicted and true test
+# signals. Store the results for plotting later.
+srm_error = score_voxelwise(
+    target_test, aligned_pred, masker=roi_masker, loss="corr"
+)
+srm_score = roi_masker.inverse_transform(srm_error)
+titles.append("Correlation of prediction after SRM alignment")
+aligned_scores.append(srm_score)
+
+# Plot the results
```
emdupre, line 217: To ease section skimming : 

```suggestion
################################################################################
# Plot the results
# -----------------
```

examples/plot_alignment_methods_benchmark.py
```diff
@@ -156,18 +179,61 @@
 
     # plot correlation for each method
     aligned_score = roi_masker.inverse_transform(method_error)
```
emdupre, line 181: I would remove this comment and move this line of code just under the next comment !

examples/plot_alignment_methods_benchmark.py
```diff
@@ -125,29 +138,39 @@
...
 #       while respecting the mass conservation constraints. Calculated with
 #       entropic regularization.
-#   *  we also include identity (no alignment) as a baseline.
-# Then for each method we define the estimator fit it, predict the new image and plot
+#   *  the shared response model (SRM), which computes a shared response space
+#      from different subjects, and then projects individual subject data into it.
+# Then for each method we define the estimator, fit it, predict the new image and plot
 # its correlation with the real signal.
 
+
```
emdupre, line 149:
```suggestion
################################################################################
# Fit and Score the Orthogonal and Optimal Transport estimators
# ---------------------------------------------------------------
```

examples/plot_alignment_methods_benchmark.py
```diff
@@ -21,14 +21,28 @@
...
+# * Emotion processing: judging facial expressions
+# * Reward processing: responding to gains and losses
+# * Working memory: maintaining sequences of faces and objects
+# * Object categorization: matching and comparing visual stimuli
+
 # We download the images for subjects sub-01 and sub-02.
-# Files is the list of paths for each subjects.
+# files is the list of paths for each subjects.
 # df is a dataframe with metadata about each of them.
+# mask is the common mask for IBC subjects.
```
emdupre, line 40: Since we're touching these lines anyway :

```suggestion
# ``files`` is the list of paths for each subjects.
# ``df`` is a dataframe with metadata about each of them.
# ``mask`` is the common mask for IBC subjects.
```

examples/plot_alignment_methods_benchmark.py
```diff
@@ -125,29 +138,39 @@
...
-for method in methods:
-    alignment_estimator = PairwiseAlignment(
-        alignment_method=method,
-        clustering="ward",
-        n_pieces=n_pieces,
-        masker=roi_masker,
+# The IdentifiableFastSRM version of SRM ensures that the solution is unique.
+#
+
+from fastsrm.identifiable_srm import IdentifiableFastSRM
```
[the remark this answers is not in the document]
emdupre: Not an important change, I just don't (personally) think we need quite all of this whitespace. Feel free to reject.

ferponcem:
Sorry about the thousand commits, I started accepting before I realized I could've done most of them in just one. 
Thank you for the great suggestions, let me know what you think about the current state ! 🚀

emdupre:
This looks great, merging now ! Thank you, @ferponcem !!!

PR #156, opened 2025-07-08
[API] GroupAlignment class

pbarbarant:
TODO:

In methods:
-̶ ̶[̶ ̶]̶ ̶A̶d̶d̶ ̶s̶r̶m̶ ̶t̶e̶s̶t̶s̶ -> We won't include SRM in this PR before #158 is solved.

In alignment.py:
- [x] Write a Piecemodel for parcellated data
- [x] Write a wrapper to compute the piecewise alignments for each subject
- [x] Write a wrapper to compute the whole-brain alignments for each subject
- [x] Write the transform method

In embeddings (maybe a different a branch):
- [x] Write the embeddings -> For another PR


examples/plot_alignment_simulated_2D_data.py
```diff
@@ -172,7 +172,7 @@ def _plot_distributions_and_alignment(
 # source `X` and the target `Y` that will handle that. Now we will showcase
 # on our simple 2D example different kind of transformation we can look for.
 
-from fmralign.alignment_methods import (
+from fmralign.methods.alignment_methods import (
```
bthirion, line 175: `fmralign.methods.alignment_methods` seems a bit redundant

fmralign/alignment.py
```diff
@@ -0,0 +1,136 @@
+from sklearn.utils.validation import check_is_fitted
+from sklearn.base import BaseEstimator, TransformerMixin
+
+from fmralign._utils import _check_labels
+import numpy as np
+from fmralign.template.utils import _rescaled_euclidean_mean
+
+
+def fit_to_target(X, target_data, method, labels, n_jobs, verbose):
+    n_labels = len(np.unique(labels))
```
bthirion, line 10: Add at least mini-docstrings asap to ease code reading.

fmralign/alignment.py
```diff
@@ -0,0 +1,136 @@
...
+        ]
+
+    # Make inherited function harmless
+    def fit_transform(self):
+        """Parent method not applicable here.
+
+        Will raise AttributeError if called.
+        """
+        raise AttributeError(
+            "type object 'PairwiseAlignment' has no attribute 'fit_transform'"
```
bthirion, line 112:
```suggestion
            "type object 'PairwiseAlignment' has no 'fit_transform' method"
```

bthirion (commented):
To be discussed, but feels OK at first sight.

bthirion:
LMK when you want another review. Best,

fmralign/alignment/group_alignment.py
```diff
@@ -0,0 +1,203 @@
...
+
+    Parameters
+    ----------
+    method : str or a `BaseAlignment` instance, default="identity"
+        The alignment method to use. Supported methods depend on the
+        underlying alignment implementation.
+    target : array-like or None, default=None
+        Target for alignment. If None, performs template alignment where
+        a template is computed from all subjects. If array-like, performs
+        pairwise alignment to the specified target data.
```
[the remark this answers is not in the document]
emdupre: The ValueError raised below makes me think that, in fact, we're expecting the reference subject id, but the example in the docstring suggests the reference subject data. We should harmonize these

emdupre (commented):
Thanks, @pbarbarant ! I think this new structure is a big improvement.

I have one core question on how we position `GroupAlignment` : from my (poor) memory, I thought we were renaming `TemplateAlignment` as `GroupAlignment` but keeping the structure the same. Here, `GroupAlignment` is effectively a wrapper to ease user interactions with either `TemplateAlignment` or `PairwiseAlignment`.

This structure is really quite elegant for the user (in that switching between the two use cases is as simple as changing the target), but I'm struggling with the conceptual distinction. That is, it seems obvious when a user would call `PairwiseAlignment` outside of `GroupwiseAlignment` (i.e., with only one target and one source subject), but it's not obvious to me when they'd call `TemplateAlignment` outside of `GroupwiseAlignment`. So, are they effectively interchangeable, but `GroupwiseAlignment` has this nice additional property of computing all pairwise alignments if requested ?

I think that makes the positioning of `TemplateAlignment` a little more complicated, and I'm left wondering if this structure is in fact a little _too_ elegant (in that isn't didactic for the users). 

This could be as simple as tweaking the naming / documentation to highlight `Pairwise` vs `Template`, and then `Groupwise` as a wrapper on both. But with the current naming, I'd expect `Groupwise` and `Pairwise` to be at the same level, and I'm left with `Template` as the odd-man-out.

doc/modules/reference.rst
```diff
@@ -67,6 +67,6 @@ uses.
 
   Identity
   DiagonalAlignment
-  ScaledOrthogonalAlignment
+  ScaledOrthogonal
```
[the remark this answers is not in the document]
pbarbarant: Good point, I'm renaming it globally.

examples/plot_alignment_methods_benchmark.py
```diff
@@ -135,7 +135,7 @@
 # its correlation with the real signal.
 
 from fmralign.metrics import score_voxelwise
-from fmralign.pairwise_alignment import PairwiseAlignment
+from fmralign.pairwise.pairwise_alignment import PairwiseAlignment
```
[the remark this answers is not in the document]
pbarbarant: I'm going to remove the examples for now. We can add them back in a final PR

fmralign/methods/ot.py
```diff
@@ -362,7 +99,7 @@ def transform(self, X):
         return X.dot(self.R)
 
 
-class SparseUOT(Alignment):
+class SparseUOT(BaseAlignment):
```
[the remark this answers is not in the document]
pbarbarant: I think we can keep it here for now, and maybe move it later as an optional [gpu] dependency.

fmralign/alignment/group_alignment.py
```diff
@@ -0,0 +1,203 @@
...
+    labels : array-like or None, default=None
+        Labels for the data features. If provided, local alignments
+        can be performed in parallel. If None, global alignment
+        is performed across all features.
+    n_jobs : int, default=1
+        Number of parallel jobs to run. -1 means using all processors.
+    verbose : int, default=0
+        Verbosity level. Higher values provide more detailed output.
+    n_iter : int, default=2
+        Number of iterations for the alignment algorithm.
```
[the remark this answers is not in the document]
pbarbarant: Yes

fmralign/_utils.py
```diff
@@ -344,7 +177,7 @@ def save_alignment(alignment_estimator, output_path):
     ValueError
         If the output path is not a valid file or directory.
     """
-    if not hasattr(alignment_estimator, "fit_"):
+    if not hasattr(alignment_estimator, "fitted_estimators"):
```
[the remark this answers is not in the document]
pbarbarant: Yes, it will be a dict with a single key/value pair accessible as `algo.fitted_estimators["sub-source"]`

examples/plot_alignment_methods_benchmark.py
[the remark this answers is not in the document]
pbarbarant: Let's make the template behavior with "template" instead of None

emdupre:
Thanks, @pbarbarant ! Confirming that my previous comment was slightly misplaced ; we had a discussion out-of-band that really helped to clarify.

Could you flag when this is ready for a final review ? 🚀

fmralign/alignment/group_alignment.py
```diff
@@ -0,0 +1,196 @@
...
+        array-like
+            Aligned subject data.
+
+        Raises
+        ------
+        ValueError
+            If the estimator has not been fitted yet.
+        """
+        # Check if estimator is fitted
+        if not hasattr(estimator, "fit"):
```
bthirion, line 147: This does not actually check that the estimator has been fitted ?

fmralign/alignment/utils.py
```diff
@@ -0,0 +1,311 @@
...
+        unique_labels, counts = np.unique(labels, return_counts=True)
+
+        if verbose > 0:
+            print(
+                f"The alignment will be applied on parcels of sizes {counts}"
+            )
+
+        if (counts > threshold).any():
+            warning = (
+                "\n Some parcels are more than 1000 voxels wide it can slow down alignment,"
```
bthirion, line 198:
```suggestion
                f"\n Some parcels are more than {threshold} voxels wide it can slow down alignment,"
```

fmralign/alignment/utils.py
```diff
@@ -0,0 +1,311 @@
...
+    target_data : ndarray
+        The target data array to which each subject's data will be fitted.
+    method : an instance of any class derived from `BaseAlignment`
+        Algorithm used to perform alignment between sources and target.
+    labels : 1D np.ndarray
+        Labels for the parcellation of the data. If only one label is present,
+        the whole brain method is used.
+        If multiple labels are present, the method will patch the parcels estimators
+        in a big whole brain estimator.
+    n_jobs : int, optional
```
bthirion, line 239:
```suggestion
    n_jobs : int, default=1
```

fmralign/alignment/utils.py
```diff
@@ -0,0 +1,311 @@
...
+        Number of jobs to run in parallel. If -1, all CPUs are used.
+        If 1, no parallel computing code is used at all, by default 1
+    verbose : int, optional
```
bthirion, line 242:
```suggestion
    verbose : int, default=0
```

fmralign/alignment/utils.py
```diff
@@ -0,0 +1,311 @@
...
+def _check_labels(X, labels=None, threshold=1000, verbose=0):
+    """Check if any parcels are bigger than set threshold.
+
+    Parameters
+    ----------
+    X : ndarray
+        The data array of shape (n_samples, n_features).
+    labels : 1D np.ndarray or None
+        Labels for the parcellation of the data.
+    threshold : int
```
bthirion, line 164:
```suggestion
    threshold : int, default=1000
```

fmralign/alignment/utils.py
```diff
@@ -0,0 +1,311 @@
...
+    X : list of 2D ndarray
+        List of subject data arrays, where each array is of shape (n_samples, n_features).
+    method : an instance of any class derived from `BaseAlignment`
+        Algorithm used to perform alignment between sources and target.
+    labels : 1D np.ndarray
+        Labels for the parcellation of the data. If only one label is present,
+        the whole brain method is used.
+        If multiple labels are present, the method will patch the parcels estimators
+        in a big whole brain estimator.
+    n_jobs : int, optional
```
bthirion, line 288:
```suggestion
    n_jobs : int, default=1
```

bthirion: etc.

fmralign/methods/procrustes.py
```diff
@@ -0,0 +1,101 @@
...
+        ||sc * RX - Y||^2 is minimized and
+        - R is an orthogonal matrix
+        - sc is a scalar
+        If scaling is false sc is set to 1
+    primal: bool or None, optional,
+         Whether the SVD is done on the YX^T (primal) or Y^TX (dual)
+         if None primal is used iff n_features <= n_timeframes
+
+    Returns
+    ----------
```
bthirion, line 29:
```suggestion
    -------
```

fmralign/methods/procrustes.py
```diff
@@ -0,0 +1,101 @@
...
+    X = X.astype(np.float64, copy=False)
+    Y = Y.astype(np.float64, copy=False)
+    if np.linalg.norm(X) == 0 or np.linalg.norm(Y) == 0:
+        return np.eye(X.shape[1]), 1
+    if primal is None:
+        primal = X.shape[0] >= X.shape[1]
+    if primal:
+        A = Y.T.dot(X)
+        if A.shape[0] == A.shape[1]:
+            A += +1.0e-18 * np.eye(A.shape[0])
```
bthirion, line 44:
```suggestion
            A += 1.0e-18 * np.eye(A.shape[0])
```

bthirion: Isn't 1.e-18 too tiny ?

fmralign/tests/alignment/test_group_alignment.py
bthirion, line 1: regarding repo organization, I think iit makes more sense to localize tests in submodules, eg
`fmralign/alignment/tests/test_group_alignment.py`
rather than
`fmralign/tests/alignment/test_group_alignment.py`

pbarbarant:
I'm finishing up sphinx doc building and it should be good to merge!

bthirion:
It caught an error.

bthirion (approved):

PR #162, opened 2025-08-21
[API] Add deterministic SRM

### LINKED ISSUE #158: [MAINT] FastSRM integration
Since we heavily rely on FastSRM which is not maintained anymore, we progressively include it inside of fmralign.

pbarbarant:
Closes #158


bthirion (approved):
The changes LGTM. IS there an example that deals with SRM atm ? If not it should be added.

examples/plot_alignment_methods_benchmark.py
```diff
@@ -133,6 +133,24 @@
 n_pieces = int(np.round(n_voxels / 100))
 print(f"We will cluster them in {n_pieces} regions")
```
bthirion, line 134: I tried to run this example but it crashed due to a dependency on fugw ??

pbarbarant: OK, we will need to cut a new release of fugw on PyPI at some point... The dependency issue is fixed for now 👍

PR #164, opened 2025-09-01
[API] Fix sphinx doc generation

pbarbarant:
Modernizes and fixes the docs to fit the new API.


examples/plot_alignment_methods_benchmark.py
```diff
@@ -138,7 +138,7 @@
 # --------------------------
 # We will use intra-parcel alignments, fmralign provides a simple utility
 # to extract the labels of associated to each voxel:
-# :func:`fmralign.embeddings.parcellation.get_labels`.
+# :func:`!fmralign.embeddings.parcellation.get_labels`.
```
emdupre, line 141: Are these not resolving correctly ?

emdupre: Hm, I can't debug now (working on slides 🙈 ) so if this fix works then let's move forward so as not to block the release. But we should likely open an issue so this doesn't stay unresolved !

PR #176, opened 2026-01-02
[DOC] Improve modules organization

pbarbarant:
Refactoring of the documentation modules following nilearn's. Adds also a `html-noplot` option to the makefile for prototyping without building the examples.


bthirion (approved):
LGTM, thx.

PR #184, opened 2026-01-19
Fix Procrustes transform

### LINKED ISSUE #183: [BUG] Procrustes transform() scales twice the transformation
`R` is already scaled before calling the method.

https://github.com/fmralign/fmralign/blob/b207768136b4bb8e3d9cfbf83bef8a3745498e1c/fmralign/methods/procrustes.py#L96-L101

pbarbarant:
Closes #183


bthirion (approved):
LGTM, thx.

PR #186, opened 2026-04-13
[FIX] Simplify OT by streamlining computations to POT

pbarbarant:
Remove SparseUOT, use torch and an option for GPU support using a device keyword. We now use POT's log-sinkhorn solver but also normalize the cost matrix for numerical stability. This is much faster than the previous sinkhorn_eps_scaling algorithm.


emdupre:
Let me know when you want a review on this !

pbarbarant:
> Let me know when you want a review on this !

It should be good now 👍

bthirion (approved):
LGTM, thx.

emdupre:
Finally had time to check : thanks, @pbarbarant  ! Happy to see this in.

PR #189, opened 2026-04-15
Add laplacian embedding utilities

pbarbarant:
Adds a utility function to get laplacian eigenmodes in volume or on a mesh.


bthirion (commented):
LGTM. You probably want to showcase it.

PR #203, opened 2026-06-24
[MAINT] test oldest dependencies

### LINKED ISSUE #202: No test with the oldest dependencies
AFAICT there is no worklfow that runs the test with the oldest supported dependencies.

I would recommend adding this to make sure everything is actually honky dory for users who may be stuck to using older versions.

Remi-Gau:
- closes #202
- run tests with oldest dependencies
- pin lowest supported version of dependencies to allow resolution by uv


pyproject.toml
```diff
@@ -11,17 +11,10 @@ authors = [
 ]
 classifiers = ["Programming Language :: Python :: 3"]
 dependencies = [
-    "numpy",
-    "pandas",
-    "matplotlib",
```
Remi-Gau, line 16: am I correct that matplotlib is only used in the doc?

Remi-Gau:
Ok I need to pin some more dependencies.

pyproject.toml
pbarbarant: Yes

Remi-Gau: Ok so I was right to move it to the doc dependencies

pyproject.toml
```diff
@@ -11,17 +11,10 @@ authors = [
...
-    "scikit-learn",
-    "joblib",
-    "scipy",
-    "nibabel",
     "nilearn>=0.13.0",
-    "POT",
-    "robust_laplacian",
-    "torch"
```
Remi-Gau: forgot to mention but a lot of this could be simplified because they are transitive dependencies via nilearn

pbarbarant:
Can you ping me once this PR is ready for a round of reviews?

bthirion (approved):

emdupre (approved):
LGTM, thx !

Remi-Gau:
@pbarbarant this is ready for review

.github/workflows/testing.yml
```diff
@@ -46,3 +46,28 @@ jobs:
...
+
+  test_min:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.10"]
+
+    steps:
+      - uses: actions/checkout@v6
```
pbarbarant:
```suggestion
      - uses: actions/checkout@v7
```

.github/workflows/testing.yml
```diff
@@ -46,3 +46,28 @@ jobs:
...
+
+      - name: Install uv and set the python version
+        uses: astral-sh/setup-uv@v7
+        with:
+          python-version: ${{ matrix.python-version }}
```
pbarbarant: Can we skip the matrix since we only run this workflow on one version?

pbarbarant (approved):
LGTM, some minor comments below:

pbarbarant:
Waiting for CI to finish then merging...

PR #216, opened 2026-07-06
[FEAT] Add geomloss backend to SpectralOT

pbarbarant:
Adds geomloss support for the SpectralOT solver. No API change apart from a new backend keyword is involved.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,11 @@ doc = [
     "furo"
 ]
 # For the ot solvers
-ot = ["geomloss[full]>=0.2.0", "POT", "torch"]
+ot = ["fmralign[ot_geomloss,ot_pot]"]
+# geomloss backend
+ot_geomloss = ["geomloss[full]>=0.2.0", "torch"]
+# POT backend
+ot_pot = ["POT", "torch"]
 # For running unit and docstring tests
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -47,6 +45,8 @@ doc = [
     "sphinxext-opengraph",
     "furo"
 ]
+# For the ot solvers
```
emdupre, line 48: Do you think we should have it s.t. if someone wants OT they have to grab all possible backends ? Or should we instead separate it out with something like :

```
ot_backends = [
    {include-group = "geomloss"},
    {include-group = "POT"},
]

# geomloss backend
geomloss_backend = ["geomloss[full]>=0.2.0", "torch"]
# POT backend
pot_backend = ["POT", "torch"]
```

pbarbarant: The elephant in the room is torch, which is common to both, but this might be better for modularity.

emdupre: It looks like POT is ~10x the size of geomloss on PyPi ; both are _much_ smaller than torch (the elephant in the room in multiple senses 😄 ), but I'm not sure that a user needs both for standard OT analyses. This would, though, mean updating the OT code to check for multiple backends rather than assume a default

bthirion (commented):
LGTM. I think it's important to give users guidance on what backend to use in the documantation.

PR #222, opened 2026-07-27
[ENH] add PiecewiseAlignment to the public API

### LINKED ISSUE #214: [TEST] Test PiecewiseAlignment sklearn estimator compliance
PiecewiseAlignment is not in the public API so I was not yet sure it is worth testing it yet: otherwise I can add it to the doc and the public API

_Originally posted by @Remi-Gau in https://github.com/fmralign/fmralign/pull/201#discussion_r3480313388_

Remi-Gau:
- closes #214 
- adds PiecewiseAlignment to the public API

### DIFF
--- a/fmralign/methods/piecewise.py
+++ b/fmralign/methods/piecewise.py
@@ -116,7 +116,7 @@ class PiecewiseAlignment(BaseAlignment):
         Verbosity level. Default is 0.
     """
 
-    def __init__(self, method, labels, n_jobs=1, verbose=0):
+    def __init__(self, method=None, labels=None, n_jobs=1, verbose=0):
         super().__init__()
         self.n_jobs = n_jobs
         self.method = method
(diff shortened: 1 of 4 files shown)


fmralign/methods/piecewise.py
```diff
@@ -133,6 +132,11 @@ def fit(self, X, Y):
         Y : np.ndarray
             Target data of shape (n_samples, n_features).
         """
+        if self.labels is None:
+            raise ValueError("'labels' cannot be None.")
```
Remi-Gau, line 136: if we cannot give some dummy values to labels based on the input data then we may have to skip a bunch of sklearn checks

emdupre: the simplest dummy value is to make everything the same value I would think ; then piecewise collapses to a single parcel

pbarbarant: An array of ones supported on the voxels is usually a good default.

PR #224, opened 2026-07-28
[MAINT] Compliance with RF101 and RF103

pbarbarant:
Turns out B and UP rules must be included in select instead of extend-select. This matches nilearn's toml.


emdupre (approved):
Thx for the fix !

